PR #3, opened 2023-02-15
Complete test for contract events

0xdev22:
# Proposed Changes


elffjs (changes requested):
This is a great start. What do you think about the following cases:

* Ignoring events that have the wrong name.
* Updating a `(contract, token, privilege, user)` combination with a new expiry, in particular looking at whether the timestamps act correctly.

internal/services/contracts_events_consumer_test.go
```diff
@@ -169,6 +172,7 @@ func TestUpdatedTimestamp(t *testing.T) {
 	}
 
 	s.assert.Equal(expected, actual, "Event was updated successful")
+	s.assert.NotEqual(oldNft.UpdatedAt, newNft.UpdatedAt)
```
elffjs, line 175: Could we check that the creation timestamp doesn't change?

elffjs (changes requested):
Just one suggestion.

I think we could spend an endless amount of time on readability but these are quality tests.

internal/services/contracts_events_consumer.go
```diff
@@ -137,7 +137,9 @@ func (c *ContractsEventsConsumer) setPrivilegeHandler(e *ContractEventData) erro
 
 	nftCols := models.NFTPrivilegeColumns
 
-	err = udp.Upsert(context.Background(), c.db.DBS().Writer, true, []string{nftCols.ContractAddress, nftCols.TokenID, nftCols.Privilege, nftCols.UserAddress}, boil.Infer(), boil.Infer())
+	err = udp.Upsert(context.Background(), c.db.DBS().Writer, true,
+		[]string{nftCols.ContractAddress, nftCols.TokenID, nftCols.Privilege, nftCols.UserAddress},
+		boil.Whitelist(nftCols.UpdatedAt), boil.Infer())
```
elffjs, line 142: Wouldn't we also want to update expiry? I'm a little confused about how this is working.

internal/services/contracts_events_consumer_test.go
```diff
@@ -0,0 +1,256 @@
...
+	s.assert.NoError(err)
+
+	args := factoryResp.args
+
+	oldNft, err := models.FindNFTPrivilege(s.ctx, s.pdb.DBS().Reader, args.contract.Bytes(), args.tokenID, args.privilegeID, args.userAddress.Bytes())
+	s.assert.NoError(err)
+
+	s.assert.NotNil(oldNft)
+
+	expiry := time.Now().Add(time.Hour + time.Duration(4)).UTC().Unix()
```
elffjs, line 141: Okay, I think I see the problem—none of these timestamps are mattering. The unit of `time.Duration` is  nanosecond, and `.Unix()` gets you to seconds.

The helper is making it very difficult to understand what's happening, but if you print out `string(msg.Payload)` in both instances you should see the same value for `expiry`.

internal/services/contracts_events_consumer.go
elffjs: I think I'm convinced that this needs to be expanded to include `nftCols.Expiry`.

elffjs (changes requested):

internal/services/contracts_events_consumer.go
elffjs: You changed the whitelist! But I think the test is still not checking anything meaningful with regard to the expiry. Let’s talk in the morning.

internal/services/contracts_events_consumer_test.go
elffjs: Yes, the scary thing is that if I remove `cold.Expiry` from the whitelist, this still passes!

internal/services/contracts_events_consumer_test.go
```diff
@@ -0,0 +1,264 @@
...
+	if eventName == "" {
+		eventName = "PrivilegeSet"
+	}
+
+	for i := from; i <= to; i++ {
+		contractAddr := common.BytesToAddress([]byte{uint8(i)})
+		userAddr := common.BytesToAddress([]byte{uint8(i + 1)})
+		tokenID := convertTokenIDToDecimal(fmt.Sprint(i))
+		privID := i + 1
+		expiry := time.Now().Add(time.Hour + time.Duration(i)).UTC().Unix()
```
elffjs, line 206: Changing this to
```suggestion
		expiry := time.Now().Add(time.Hour + time.Duration(i)*time.Minute).UTC().Unix()
```
makes the whitelist matter.

elffjs (approved):

PR #4, opened 2023-02-15
Main backup (cherry picked commits in main_backup from origin/create_autopi_templates)

Khalaro:
Testing still needed

# Proposed Changes


cmd/devices-api/main.go
```diff
@@ -245,6 +246,39 @@ func main() {
 			logger.Fatal().Err(err).Msg("failed to sync all devices with their templates")
 		}
 		logger.Info().Msg("success")
+		//							1				2		3		4			5
+	case "autopi-tools": //   autopi-tools   templateName  [-p  parent]  description
+		var newTemplateIndex int
```
JamesReate, line 251: let's move the contents of this to a new file in `cmd/devices-api` eg. `autopi-tools.go` and then just call a function from there. That way we can keep the logic separated from the entrypoint.

internal/services/autopi_api_service.go
```diff
@@ -346,6 +390,21 @@ type postDeviceIDs struct {
...
+// used to create a new AutoPi template on the cloud
+type postNewTemplateRequest struct {
+	TemplateName string   `json:"templateName"`
+	Parent       int      `json:"parent,omitempty"`
+	Description  string   `json:"description"`
+	Devices      []string `json:"devices"`
+}
+
+// response structure after creating a new AutoPi template on the cloud
+type postNewTemplateResponse struct {
```
JamesReate, line 402: looks like this isn't being used - remove if don't intend to use.

JamesReate (commented):
you'll notice a line at the top:
`//go:generate mockgen -source autopi_api_service.go -destination mocks/autopi_api_service_mock.go`
you'll need to run it to regenerate the test mocks for the interface. In Goland it's pretty easy, just click on the play button next to it and select the 3rd option (last one)

JamesReate (commented):
To run the linter locally: 
install on windows:
https://golangci-lint.run/usage/install/#linux-and-windows

then from terminal
golangci-lint run

PR #6, opened 2023-02-17
Consumer for mint aftermarket devices

0xdev22:
# Proposed Changes


internal/services/contracts_events_consumer.go
```diff
@@ -145,3 +153,51 @@ func (c *ContractsEventsConsumer) setPrivilegeHandler(e *ContractEventData) erro
 
 	return nil
 }
+
+type MintedAftermarketDeviceArgs struct {
+	TokenID       string `mapstructure:"newTokenId"`
+	DeviceAddress string
+	Owner         string // TODO - confirm who the minter is
```
elffjs, line 160: I'm a little more worried about checking that the contract is what we think it should be. The owner is a bit of a weird thing here.

internal/services/contracts_events_consumer.go
```diff
@@ -145,3 +153,51 @@ func (c *ContractsEventsConsumer) setPrivilegeHandler(e *ContractEventData) erro
 
 	return nil
 }
+
+type MintedAftermarketDeviceArgs struct {
```
elffjs, line 157: It would feel good to grab this from the ABI but I don't think that'll play well with `mapstructure`.

internal/services/contracts_events_consumer.go
```diff
@@ -145,3 +153,51 @@ func (c *ContractsEventsConsumer) setPrivilegeHandler(e *ContractEventData) erro
...
+	autopiApiService := NewAutoPiAPIService(c.settings, c.db.DBS)
+	device, err := autopiApiService.GetDeviceByEthAddress(p.DeviceAddress)
+	if err != nil {
+		c.log.Error().Msg(fmt.Sprintf("Couldn't fetch dongle with eth_address %s.", p.DeviceAddress))
+		return fmt.Errorf("couldn't fetch dongle with eth_address %s", p.DeviceAddress)
+	}
+
+	var maybeAddr null.Bytes
+
+	if strAddr := device.EthereumAddress; common.IsHexAddress(p.DeviceAddress) {
```
elffjs, line 179: If it's not an address then I would expect the call on 171 to fail; I doin't think you can mint without an address at this point. Maybe we check earlier than this?

internal/services/contracts_events_consumer.go
```diff
@@ -145,3 +153,51 @@ func (c *ContractsEventsConsumer) setPrivilegeHandler(e *ContractEventData) erro
...
+	ap := models.AutopiUnit{
+		AutopiUnitID:    device.UnitID,
+		AutopiDeviceID:  null.StringFrom(device.ID),
+		EthereumAddress: maybeAddr,
+		TokenID:         types.NewNullDecimal(new(decimal.Big).SetBigMantScale(tokenID, 0)),
+	}
+
+	apCols := models.AutopiUnitColumns
+
+	err = ap.Upsert(context.Background(), c.db.DBS().Writer, true, []string{apCols.AutopiUnitID, apCols.EthereumAddress}, boil.Infer(), boil.Infer())
```
elffjs, line 196: I'm a little worried about what this will do if there's an existing record with, e.g., `user_id` populated. I never remember what it is that `boil.Infer` does.

internal/services/contracts_events_consumer.go
```diff
@@ -145,3 +153,51 @@ func (c *ContractsEventsConsumer) setPrivilegeHandler(e *ContractEventData) erro
...
+
+	var maybeAddr null.Bytes
+
+	if strAddr := device.EthereumAddress; common.IsHexAddress(p.DeviceAddress) {
+		maybeAddr = null.BytesFrom(common.FromHex(strAddr))
+	} else {
+		c.log.Warn().Str("address", device.EthereumAddress).Msg("Invalid device Ethereum address from AutoPi.")
+	}
+
+	tokenID, _ := new(big.Int).SetString(p.TokenID, 16)
```
elffjs, line 185: Is it in base 16? That would be somewhat surprising to me. I can show you how to play with some of the real output from contract-event-processor. This is probably the most serious comment here.

If we're going to check that the address is valid we might as well make sure that this works.

We can probably bypass `big.Int` and go straight to `decimal.Big`.

internal/services/autopi_api_service.go
```diff
@@ -108,6 +109,23 @@ func (a *autoPiAPIService) GetDeviceByID(deviceID string) (*AutoPiDongleDevice,
 	return d, nil
 }
 
+// GetDeviceByEthAddress calls https://api.dimo.autopi.io/dongle/devices/by_eth_address/{eth_address}/. This helps get the device for the eth_address.
+// Throws an error if none or more than one is found.
+func (a *autoPiAPIService) GetDeviceByEthAddress(ethAddress string) (*AutoPiDongleDevice, error) {
```
elffjs, line 114: 👍

internal/services/contracts_events_consumer.go
```diff
@@ -145,3 +153,51 @@ func (c *ContractsEventsConsumer) setPrivilegeHandler(e *ContractEventData) erro
...
+	p := MintedAftermarketDeviceArgs{}
+	err := mapstructure.WeakDecode(e.Arguments, &p)
+	if err != nil {
+		return err
+	}
+
+	autopiApiService := NewAutoPiAPIService(c.settings, c.db.DBS)
+	device, err := autopiApiService.GetDeviceByEthAddress(p.DeviceAddress)
+	if err != nil {
+		c.log.Error().Msg(fmt.Sprintf("Couldn't fetch dongle with eth_address %s.", p.DeviceAddress))
```
elffjs, line 173: I think the error on the next line will get logged anyway.

elffjs (changes requested):
Just one serious comment and some minor suggestions. Looks great.

internal/services/contracts_events_consumer.go
0xdev22: I guess the question then would be what fields could possibly change to require an insert. Seems the fields being passed from the contract events listener are mostly identifiers (ID's) which to me doesn't seem like something that will change easily. So do we just do an `insert` only here instead of an `upsert`

internal/services/autopi_api_service.go
```diff
@@ -108,6 +109,23 @@ func (a *autoPiAPIService) GetDeviceByID(deviceID string) (*AutoPiDongleDevice,
...
+	res, err := a.httpClient.ExecuteRequest(fmt.Sprintf("/dongle/devices/by_eth_address/%s/", ethAddress), "GET", nil)
```
elffjs, line 115: I just checked and the casing of the address doesn't matter, that's good news.

internal/services/contracts_events_consumer.go
elffjs: My question is about devices that are already in our system but not minted. These days such devices are usually devices that need to generate a private key in software after getting the appropriate software update. They may have fields assigned in the database like `user_id`, but they lack an `ethereum_address` and a `token_id`.

My worry is that SQLBoiler here will blow away those fields when the minting message comes in.

elffjs: I guess the basic problem is that I don't think SQLBoiler can distinguish between "you should update this to null" and "you should leave this field alone".

internal/services/contracts_events_consumer.go
elffjs: On that point, these field names don't look quite right.

elffjs (approved):
Rewrote a chunk to get this moving while AutoPi is here. If you think it stinks please tell me!

PR #32, opened 2023-03-14
PLA-1092 Error code query storage and API

0xdev22:
# Proposed Changes


elffjs:
Would you fix the merge conflict?

elffjs (changes requested):
Only serious qualms:

- Cascading
- Sorting by and including times

internal/controllers/device_data_controller.go
```diff
@@ -290,26 +302,63 @@ func (udc *UserDevicesController) QueryDeviceErrorCodes(c *fiber.Ctx) error {
...
 
 	return c.JSON(&QueryDeviceErrorCodesResponse{
 		Message: chtResp,
 	})
 }
 
+// GetUserDevicesErrorCodeQueries godoc
+// @Description Returns all error codes queries for user devices
+// @Tags        user-devices
+// @Success     200 {object} controllers.GetUserDevicesErrorCodeQueriesResponse
```
elffjs, line 327: Is this accurate?

internal/controllers/device_data_controller.go
```diff
@@ -290,26 +302,63 @@ func (udc *UserDevicesController) QueryDeviceErrorCodes(c *fiber.Ctx) error {
...
+			ud = append(ud, GetUserDevicesErrorCodeQueriesResponse{
+				Codes:       erc.ErrorCodes,
+				Description: erc.QueryResponse,
+			})
+		}
+
+		resp[userDevice.ID] = ud
+	}
+
+	return c.JSON(resp)
```
elffjs, line 359: I forgot something important: I would not return a map like this at the top level. I think it constrains what you can add to the output later too much.

internal/controllers/device_data_controller.go
```diff
@@ -28,6 +30,12 @@ type QueryDeviceErrorCodesResponse struct {
 	Message string `json:"message"`
 }
 
+type GetUserDevicesErrorCodeQueriesResponse struct {
+	Codes       []string
+	Description string
+	RequestedAt time.Time
```
elffjs, line 36: I think we should JSON tag these to match the rest of the API.

internal/controllers/device_data_controller.go
```diff
@@ -318,11 +326,55 @@ func (udc *UserDevicesController) QueryDeviceErrorCodes(c *fiber.Ctx) error {
...
+// @Tags        user-devices
+// @Success     200 {object} controllers.GetUserDevicesErrorCodeQueriesResponse
+// @Security    BearerAuth
+// @Router      /user/devices/error-codes [get]
+func (udc *UserDevicesController) GetUserDevicesErrorCodeQueries(c *fiber.Ctx) error {
+	userID := helpers.GetUserID(c)
+
+	userDevices, err := models.UserDevices(
+		models.UserDeviceWhere.UserID.EQ(userID),
+		qm.Load(models.UserDeviceRels.ErrorCodeQueries),
```
elffjs, line 353: I think we need to put the order by inside of here. I think we're ordering the _devices_ here but not the queries.

elffjs (changes requested):

PR #37, opened 2023-03-17
PLA-1156 Remove year from ChatGPT query and update tests

0xdev22:
# Proposed Changes

### DIFF
--- a/internal/controllers/device_data_controller.go
+++ b/internal/controllers/device_data_controller.go
@@ -271,7 +271,7 @@ func (udc *UserDevicesController) RefreshUserDeviceStatus(c *fiber.Ctx) error {
 	return fiber.NewError(fiber.StatusBadRequest, "no active Smartcar integration found for this device")
 }
 
-var errorCodeRegex = regexp.MustCompile(`^[A-Z0-9]{5,8}$`)
+var errorCodeRegex = regexp.MustCompile(`^.{5,8}$`)
 
 // QueryDeviceErrorCodes godoc
 // @Description Queries chatgpt for user device error codes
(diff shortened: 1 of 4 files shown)


internal/controllers/device_data_controller.go
```diff
@@ -271,7 +271,7 @@ func (udc *UserDevicesController) RefreshUserDeviceStatus(c *fiber.Ctx) error {
 	return fiber.NewError(fiber.StatusBadRequest, "no active Smartcar integration found for this device")
 }
 
-var errorCodeRegex = regexp.MustCompile(`^[A-Z0-9]{5,8}$`)
+var errorCodeRegex = regexp.MustCompile(`^.{5,8}$`)
```
elffjs, line 274: This seems like the second major change here, removing validation outside of length.

internal/services/openai.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 type OpenAI interface {
-	GetErrorCodesDescription(make, model string, year int32, errorCodes []string) (string, error)
+	GetErrorCodesDescription(make, model string, errorCodes []string) (string, error)
```
elffjs, line 17: I don't feel super strongly about this, but if you look at [the other services](https://github.com/DIMO-Network/devices-api/blob/9dc5829651b1000b1c112881df84b5b0a53562cb/internal/services/autopi_api_service.go#L25) they put a `//go:generate` comment at the top. Then `go generate ./...` builds all the mocks.

internal/services/openai.go
```diff
@@ -89,16 +89,16 @@ func (o openAI) askChatGPT(body io.Reader) (*ChatGPTResponse, error) {
...
-func (o openAI) GetErrorCodesDescription(make, model string, year int32, errorCodes []string) (string, error) {
+func (o openAI) GetErrorCodesDescription(make, model string, errorCodes []string) (string, error) {
 	codes := strings.Join(errorCodes, ", ")
 	req := fmt.Sprintf(`{
 		"model": "gpt-3.5-turbo",
 		"messages": [
 			{
 				"role": "user", 
-				"content": "Briefly summarize for me, in order, what the following error codes mean for a %s %s %d. The error codes are %s."}]
+				"content": "Briefly summarize for me, in order, what the following error codes mean for a %s %s. The error codes are %s."}]
```
elffjs, line 99: I could have sworn I fixed up the formatting here. I guess it got undone by the second PR.

elffjs: More importantly: the last `%s` might be replaced by any characters. `"`, `"}`, etc. It's not the worst hole in the world but it seems needless.

elffjs (approved):

elffjs:
Another thing we might worry about: new models. If we ask it about a Cadillac Lyriq, for example, what is it going to say?

PR #38, opened 2023-03-18
[EXP-402] add growth percentage calculation on user device valuation

Asfiroth:
# Proposed Changes
- add growth percentage to user devices valuation


JamesReate (approved):
let's see if can resolve some of the simpler linter warnings like:
`unused-parameter: parameter 'empty' seems to be unused, consider removing or renaming it as _ (revive)`

and also see why that test is failing

PR #40, opened 2023-03-20
PLA-1157 Change route for fetching error codes

0xdev22:
# Proposed Changes


internal/controllers/device_data_controller.go
```diff
@@ -344,12 +344,14 @@ func (udc *UserDevicesController) QueryDeviceErrorCodes(c *fiber.Ctx) error {
...
-// @Router      /user/devices/error-codes [get]
+// @Router      /user/devices/{userDeviceID}/error-codes [get]
 func (udc *UserDevicesController) GetUserDevicesErrorCodeQueries(c *fiber.Ctx) error {
+	udi := c.Params("userDeviceID")
 	userID := helpers.GetUserID(c)
 
 	userDevices, err := models.UserDevices(
 		models.UserDeviceWhere.UserID.EQ(userID),
+		models.UserDeviceWhere.ID.EQ(udi),
 		qm.Load(models.UserDeviceRels.ErrorCodeQueries, qm.OrderBy("created_at desc")),
```
elffjs, line 355:
```suggestion
		qm.Load(models.UserDeviceRels.ErrorCodeQueries, qm.OrderBy(models.ErrorCodeQueryColumns.CreatedAt + " DESC")),
```
Minor suggestion.

elffjs (changes requested):

internal/controllers/device_data_controller.go
```diff
@@ -344,12 +344,14 @@ func (udc *UserDevicesController) QueryDeviceErrorCodes(c *fiber.Ctx) error {
...
 	).All(c.Context(), udc.DBS().Reader)
```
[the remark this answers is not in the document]
0xdev22: Using .`One` here only returns one of the stored queries for the userDevice and since we can have multiple error codes it won't work in this case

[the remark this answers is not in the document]
elffjs: The top level table is `user_devices`. I don’t think `.One` is going to constrain the load on the relation.

cmd/devices-api/main.go
```diff
@@ -77,12 +77,12 @@ func main() {
 
 	// Run API
 	if len(os.Args) == 1 {
-		if settings.EnablePrivileges {
+		/* if settings.EnablePrivileges {
```
elffjs, line 80: Is this left behind from local debugging? Could we get rid of it?

internal/controllers/device_data_controller.go
```diff
@@ -344,31 +344,30 @@ func (udc *UserDevicesController) QueryDeviceErrorCodes(c *fiber.Ctx) error {
...
+		qm.Load(models.UserDeviceRels.ErrorCodeQueries, qm.OrderBy(models.ErrorCodeQueryColumns.CreatedAt+" DESC")),
 	).All(c.Context(), udc.DBS().Reader)
 	if err != nil {
 		return err
 	}
 
-	resp := map[string][]GetUserDevicesErrorCodeQueriesResponse{}
+	resp := []GetUserDevicesErrorCodeQueriesResponse{}
 
 	for _, userDevice := range userDevices {
```
elffjs, line 363: I think that we’re iterating over a list of vehicles for a one-vehicle endpoint should give us pause!

internal/controllers/device_data_controller.go
```diff
@@ -344,31 +344,30 @@ func (udc *UserDevicesController) QueryDeviceErrorCodes(c *fiber.Ctx) error {
 // @Tags        user-devices
 // @Success     200 {object} controllers.GetUserDevicesErrorCodeQueriesResponse
 // @Security    BearerAuth
-// @Router      /user/devices/error-codes [get]
+// @Router      /user/devices/{userDeviceID}/error-codes [get]
 func (udc *UserDevicesController) GetUserDevicesErrorCodeQueries(c *fiber.Ctx) error {
```
elffjs, line 348: Oh maybe we should change the function name too—`GetUserDeviceErrorCodeQueries`?

internal/controllers/device_data_controller.go
```diff
@@ -344,31 +344,30 @@ func (udc *UserDevicesController) QueryDeviceErrorCodes(c *fiber.Ctx) error {
...
 				Codes:       erc.ErrorCodes,
 				Description: erc.QueryResponse,
 				RequestedAt: erc.CreatedAt,
 			})
 		}
-
-		resp[userDevice.ID] = ud
 	}
 
 	return c.JSON(resp)
```
elffjs, line 373: Could we put this list in a wrapper object? Just in case.

internal/controllers/device_data_controller_test.go
```diff
@@ -611,14 +611,14 @@ func TestUserDevicesController_GetUserDevicesErrorCodeQueries(t *testing.T) {
...
 
-		request := test.BuildRequest("GET", "/user/devices/error-codes", "")
+		request := test.BuildRequest("GET", fmt.Sprintf("/user/devices/%s/error-codes", ud.ID), "")
 		response, _ := app.Test(request)
 		body, _ := io.ReadAll(response.Body)
 
 		assert.Equal(t, fiber.StatusOK, response.StatusCode)
 
 		assert.JSONEq(t,
-			fmt.Sprintf(`{"%s":[{"errorCodes":["P0017","P0016"],"description":"1. P0113 - Engine Coolant Temperature Circuit Malfunction: This code indicates that the engine coolant temperature sensor is sending a signal that is outside of the expected range, which may cause the engine to run poorly or overheat.", "requestedAt":"%s"}]}`, ud.ID, currTime.UTC().Truncate(time.Microsecond).Format(time.RFC3339Nano)),
```
elffjs, line 621: Hmm, I wonder why this is failing.

elffjs (changes requested):

PR #41, opened 2023-03-20
PLA-1158 Add cascade delete to error codes

0xdev22:
# Proposed Changes

### DIFF
--- a/migrations/20230313102015_error_code_queries.sql
+++ b/migrations/20230313102015_error_code_queries.sql
@@ -10,7 +10,7 @@ CREATE TABLE devices_api.error_code_queries(
     updated_at           timestamptz not null default current_timestamp,
 
     PRIMARY KEY (id),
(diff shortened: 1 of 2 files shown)


migrations/20230313102015_error_code_queries.sql
```diff
@@ -10,7 +10,7 @@ CREATE TABLE devices_api.error_code_queries(
     updated_at           timestamptz not null default current_timestamp,
 
     PRIMARY KEY (id),
-    CONSTRAINT error_code_queries_user_device_id_fkey FOREIGN KEY (user_device_id) REFERENCES user_devices(id)
+    CONSTRAINT error_code_queries_user_device_id_fkey FOREIGN KEY (user_device_id) REFERENCES user_devices(id) ON DELETE CASCADE
```
elffjs, line 13: Unfortunately we have already pushed this migration to dev and prod. You will need to add a new migration that drops the constraint and re-adds it.

elffjs (changes requested):

elffjs (approved):

PR #42, opened 2023-03-22
PLA-1118 close loopholes in /fromsmartcar endpoint

JamesReate:
# Proposed Changes
- user id in redis cache key
- duplicate vin check


JamesReate:
@opula FYI - do not call the /fromsmartcar endpoint if for example the user already has an autopi integration and they're just wanting to add a smartcar integration on an existing car. Basically this endpoint should only be used for a new car (user_device). Otherwise the vin check would fail.

PR #44, opened 2023-03-28
Create OpenAI metrics and measure

0xdev22:
# Proposed Changes


internal/appmetrics/counters.go
```diff
@@ -31,4 +31,30 @@ var (
 		Name: "devices_api_blackbook_request_success_ops_total",
 		Help: "Total successful Blackbook used",
 	})
+	// Chat GPT Metrics
+	OpenAITotalCallsOps = promauto.NewCounter(prometheus.CounterOpts{
+		Name:      "devices_api_openai_chatgpt_ops_total",
```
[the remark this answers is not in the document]
elffjs: I think `devices_api_error_codes_openai_requests_total` is good, something like that.

internal/appmetrics/counters.go
```diff
@@ -31,4 +31,30 @@ var (
...
+	})
+	OpenAITotalFailedCallsOps = promauto.NewCounter(prometheus.CounterOpts{
+		Name:      "devices_api_openai_chatgpt_ops_failed_calls_total",
+		Help:      "Total number of failed calls to Open AI ChatGPT",
+		Namespace: "devices_api",
+		Subsystem: "error_codes",
+	})
+	OpenAITotalTokensUsedOps = promauto.NewCounter(prometheus.CounterOpts{
+		Name:      "devices_api_openai_chatgpt_ops_total_token_used",
+		Help:      "Total number of failed calls to Open AI ChatGPT",
```
elffjs, line 49: Different help text? To be honest I never bother with this.

internal/appmetrics/counters.go
```diff
@@ -31,4 +31,30 @@ var (
...
 	})
+	// Chat GPT Metrics
+	OpenAITotalCallsOps = promauto.NewCounter(prometheus.CounterOpts{
+		Name:      "devices_api_openai_chatgpt_ops_total",
+		Help:      "Total number of calls to Open AI ChatGPT",
+		Namespace: "devices_api",
+		Subsystem: "error_codes",
+	})
+	OpenAITotalFailedCallsOps = promauto.NewCounter(prometheus.CounterOpts{
+		Name:      "devices_api_openai_chatgpt_ops_failed_calls_total",
```
elffjs, line 42: This is super picky but `ops` doesn't seem to be doing much here. `devices_api_error_codes_openai_failed_requests_total`?

I mostly borrowed naming conventions from [here](https://prometheus.io/docs/practices/naming/) and [here](https://prometheus.io/docs/practices/instrumentation/#failures).

internal/appmetrics/counters.go
```diff
@@ -31,4 +31,30 @@ var (
...
+		Help:      "Total number of failed calls to Open AI ChatGPT",
+		Namespace: "devices_api",
+		Subsystem: "error_codes",
+	})
+	OpenAIResponseTimeOps = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:      "devices_api_openai_chatgpt_ops_request_duration_seconds",
+		Help:      "Response duration of OpenAI ChatGPT in seconds",
+		Namespace: "devices_api",
+		Subsystem: "error_codes",
+		Buckets:   []float64{0.1, 0.15, 0.2, 0.25, 0.3, 0.5, 0.7, 0.9, 10},
```
elffjs, line 58: Given what you've seen, are these reasonable buckets? I feel like I've never seen it take less than 5 seconds. Most of these buckets won't be helpful.

We probably won't get this right the first time.

elffjs: I think there are some built-in Prometheus rules that can replace all of these besides the token count.

internal/services/openai.go
```diff
@@ -89,7 +101,17 @@ func (o openAI) askChatGPT(body io.Reader) (*ChatGPTResponse, error) {
 	return cResp, nil
 }
 
-func (o openAI) GetErrorCodesDescription(make, model string, errorCodes []string) (string, error) {
+func (o *openAI) measureAPIResponseTime(req *http.Request, start time.Time) *http.Request {
+	trace := &httptrace.ClientTrace{
+		GotFirstResponseByte: func() {
+			o.currentReqResponseTime = time.Since(start)
```
elffjs, line 107: I think attaching this to the shared OpenAI client exposes us to races while not offering much benefit. The user of `httptrace` is interesting.

internal/services/openai.go
```diff
@@ -69,12 +74,19 @@ func (o openAI) askChatGPT(body io.Reader) (*ChatGPTResponse, error) {
...
+
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
 		return nil, err
 	}
 
 	defer resp.Body.Close()
+
+	appmetrics.OpenAIResponseTimeOps.With(prometheus.Labels{
+		"method": "GET",
```
elffjs, line 87: The method doesn't seem to add much.

internal/services/openai.go
```diff
@@ -18,10 +22,11 @@ type OpenAI interface {
...
 type openAI struct {
-	chatGptURL string
-	token      string
-	httpClient *http.Client
-	logger     *zerolog.Logger
+	chatGptURL             string
+	token                  string
+	httpClient             *http.Client
+	logger                 *zerolog.Logger
+	currentReqResponseTime time.Duration
```
elffjs, line 29: Let's figure out a version that doesn't have this.

elffjs (changes requested):
I wasn't expecting all these labels—I think if we're going that route we could make some use of [`promhttp`](https://pkg.go.dev/github.com/prometheus/client_golang@v1.14.0/prometheus/promhttp). Let me propose some stuff.

elffjs:
Here, how's this? https://github.com/DIMO-Network/temp-measure

internal/appmetrics/counters.go
```diff
@@ -31,4 +31,30 @@ var (
...
+		Namespace: "devices_api",
+		Subsystem: "error_codes",
+	})
+	OpenAITotalFailedCallsOps = promauto.NewCounter(prometheus.CounterOpts{
+		Name:      "devices_api_openai_chatgpt_ops_failed_calls_total",
+		Help:      "Total number of failed calls to Open AI ChatGPT",
+		Namespace: "devices_api",
+		Subsystem: "error_codes",
+	})
+	OpenAITotalTokensUsedOps = promauto.NewCounter(prometheus.CounterOpts{
```
elffjs, line 47: Oh, do we ever use this? I remember there are token counts from the input and output, and I figured we would add those numbers up and increase the counter by that much.

elffjs (changes requested):

internal/appmetrics/counters.go
0xdev22: I've seen it take 11seconds actually

0xdev22: Except we decide to give it a cutoff at say 10 and not care about anything after

internal/appmetrics/counters.go
0xdev22: There is total token count being returned as well. I have made changes to add the values to `OpenAITotalTokensUsedOps`  now

0xdev22:
> Here, how's this? https://github.com/DIMO-Network/temp-measure

This seems to be a middleware or a least is working like one.

internal/services/openai.go
```diff
@@ -69,12 +73,25 @@ func (o openAI) askChatGPT(body io.Reader) (*ChatGPTResponse, error) {
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("Authorization", "Bearer "+o.token)
 
+	var currentReqResponseTime time.Duration
+	trace := &httptrace.ClientTrace{
+		GotFirstResponseByte: func() {
+			currentReqResponseTime = time.Since(time.Now())
```
elffjs, line 79: `time.Since(time.Now())` doesn't seem right. That should be very close to zero. Maybe a variable in the enclosing function that holds the start time?

internal/services/openai.go
```diff
@@ -69,12 +73,25 @@ func (o openAI) askChatGPT(body io.Reader) (*ChatGPTResponse, error) {
...
+		},
+	}
+	req = req.WithContext(httptrace.WithClientTrace(req.Context(), trace))
```
elffjs, line 82: Good to use a context here in a general, in case we need to cancel the long-running request.

internal/services/openai.go
```diff
@@ -69,12 +73,25 @@ func (o openAI) askChatGPT(body io.Reader) (*ChatGPTResponse, error) {
...
+
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
 		return nil, err
 	}
 
 	defer resp.Body.Close()
+
+	appmetrics.OpenAIResponseTimeOps.With(prometheus.Labels{
+		"api_url": o.chatGptURL,
```
elffjs, line 92: Status seems okay, but I would just drop this `api_url` label. It never changes—if it ever does we can add some stuff later.

internal/services/openai.go
```diff
@@ -69,12 +73,25 @@ func (o openAI) askChatGPT(body io.Reader) (*ChatGPTResponse, error) {
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("Authorization", "Bearer "+o.token)
 
+	var currentReqResponseTime time.Duration
+	trace := &httptrace.ClientTrace{
+		GotFirstResponseByte: func() {
```
elffjs, line 78: I wonder if first byte is the right thing. Is first byte-to-last byte pretty fast?

internal/appmetrics/counters.go
elffjs: Much better now that it is being used.

elffjs (changes requested):

elffjs:
> > Here, how's this? https://github.com/DIMO-Network/temp-measure
> 
> This seems to be a middleware or a least is working like one.

Sure, I thought you might want to use that instead. But this is fine, too.

elffjs (approved):

internal/services/openai.go
elffjs: Do I care about first byte or last byte? Anyway, this is fine.

internal/appmetrics/counters.go
```diff
@@ -31,4 +31,22 @@ var (
...
+		Help: "Total number of failed calls to Open AI ChatGPT",
+	})
+	OpenAITotalTokensUsedOps = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "devices_api_error_codes_openai_total_token_used",
+		Help: "Total number of failed calls to Open AI ChatGPT",
+	})
+	OpenAIResponseTimeOps = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "devices_api_error_codes_openai_request_duration_seconds",
+		Help:    "Response duration of OpenAI ChatGPT in seconds",
+		Buckets: []float64{0.1, 0.15, 0.2, 0.25, 0.3, 0.5, 0.7, 0.9, 10},
```
elffjs, line 50: We should revisit these buckets at some point. I doubt `0.1` to `0.9` are ever going to get hit.

PR #45, opened 2023-03-30
PLA-1159 - aftermarket device transfer

0xdev22:
# Proposed Changes

### DIFF
--- a/charts/devices-api/values.yaml
+++ b/charts/devices-api/values.yaml
@@ -4,12 +4,12 @@ image:
   pullPolicy: IfNotPresent
   tag: fae79eb
 imagePullSecrets: []
-nameOverride: ''
-fullnameOverride: ''
+nameOverride: ""
+fullnameOverride: ""
 serviceAccount:
   create: true
   annotations: {}
-  name: ''
+  name: ""
 podAnnotations: {}
 podSecurityContext:
   runAsNonRoot: true
@@ -27,18 +27,18 @@ deployJob:
   enabled: false
 env:
   ENVIRONMENT: dev
-  PORT: '8080'
+  PORT: "8080"
   LOG_LEVEL: info
-  DB_PORT: '5432'
+  DB_PORT: "5432"
   DB_NAME: devices_api
-  DB_MAX_OPEN_CONNECTIONS: '10'
-  DB_MAX_IDLE_CONNECTIONS: '3'
+  DB_MAX_OPEN_CONNECTIONS: "10"
+  DB_MAX_IDLE_CONNECTIONS: "3"
   SERVICE_NAME: devices-api
   JWT_KEY_SET_URL: https://auth.dev.dimo.zone/keys
   DEPLOYMENT_BASE_URL: https://devices-api.dev.dimo.zone
   TOR_PROXY_URL: socks5://tor.p2p:9050
   REDIS_USER: default
-  REDIS_PORT: '6379'
+  REDIS_PORT: "6379"
   REDIS_TLS: true
   NFT_INPUT_TOPIC: topic.device.nft.mint
   NFT_OUTPUT_TOPIC: topic.device.nft.mint.result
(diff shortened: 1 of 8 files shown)


elffjs (changes requested):
Updating the owner address and using the right token id are the important bits.

Let's do some tests:

* Event from mint, skip (unfortunately `AftermarketDeviceNodeMinted` comes _after_ the `Transfer`)
* Event from claim, skip
* Event from a `.transfer` call on a claimed AutoPi: updates the right AutoPi (maybe we shouldn't use the token id 1 here) with the new owner and clears the user field.

The overall shape is good.

internal/contracts/aftermarket_device_id.go
elffjs, line 1: Regenerated this as
```
abigen --abi AftermarketDeviceId.json --pkg contracts --type AftermarketDeviceId --out aftermarket_device_id.go
```

internal/config/settings.go
```diff
@@ -74,6 +74,7 @@ type Settings struct {
 	VincarioAPIKey                    string      `yaml:"VINCARIO_API_KEY"`
 	OpenAISecretKey                   string      `yaml:"OPENAI_SECRET_KEY"`
 	ChatGPTURL                        string      `yaml:"CHATGPT_URL"`
+	AftermarketDeviceContractAddress        string      `yaml:"AFTERMARKET_DEVICE_CONTRACT_ADDRESS"`
```
elffjs, line 77: Can you update `values.yaml`? Remember to format this file.

internal/services/contracts_events_consumer.go
```diff
@@ -101,18 +103,27 @@ func (c *ContractsEventsConsumer) processEvent(event *shared.CloudEvent[json.Raw
 		return err
 	}
 
-	switch data.EventName {
-	case PrivilegeSet.String():
-		c.log.Info().Str("event", data.EventName).Msg("Event received")
-		return c.setPrivilegeHandler(&data)
-	case AftermarketDeviceNodeMinted.String():
-		if data.Contract == c.registryAddr {
+	if event.Source == fmt.Sprintf("chain/%d", c.settings.DIMORegistryChainID) {
```
elffjs, line 106: Could we do check-and-return for this exceptional case? I think it would read better.

internal/services/contracts_events_consumer.go
```diff
@@ -101,18 +103,27 @@ func (c *ContractsEventsConsumer) processEvent(event *shared.CloudEvent[json.Raw
...
+		switch data.EventName {
+		case PrivilegeSet.String():
 			c.log.Info().Str("event", data.EventName).Msg("Event received")
-			return c.setMintedAfterMarketDevice(&data)
+			return c.setPrivilegeHandler(&data)
+		case Transfer.String():
+
+			c.log.Info().Str("event", data.EventName).Msg("Event received")
+			return c.routeTransferEvent(&data)
+
```
elffjs, line 115:
```suggestion
			c.log.Info().Str("event", data.EventName).Msg("Event received")
			return c.routeTransferEvent(&data)
```

internal/services/contracts_events_consumer.go
```diff
@@ -125,8 +136,62 @@ type PrivilegeArgs struct {
...
+		c.log.Err(err).Str("tokenID", tkID.String()).Msg("Device has not been claimed yet")
+		return nil
+	}
+
+	apUnit.UserID = null.String{}
+	apUnit.OwnerAddress = null.BytesFrom(args.To.Bytes())
+
+	cols := models.AutopiUnitColumns
+
+	_, err = apUnit.Update(ctx, c.db.DBS().Writer, boil.Whitelist(cols.UserID))
```
elffjs, line 185: Remember to update the whitelist.

elffjs (changes requested):

charts/devices-api/values.yaml
```diff
@@ -67,15 +67,16 @@ env:
...
   MONITORING_SERVER_PORT: 8888
   ENABLE_PRIVILEGES: true
   TOKEN_EXCHANGE_JWK_KEY_SET_URL: http://dex-roles-rights.dev.svc.cluster.local:5556/keys
   CONTRACT_EVENT_TOPIC: topic.contract.event
-  VEHICLE_NFT_ADDRESS: '0x90c4d6113ec88dd4bdf12f26db2b3998fd13a144'
+  VEHICLE_NFT_ADDRESS: "0x90c4d6113ec88dd4bdf12f26db2b3998fd13a144"
   AUTOPI_NFT_IMAGE: autopi.png
   VINCARIO_API_URL: https://api.vindecoder.eu/3.2
   CHATGPT_URL: https://api.openai.com/v1/chat/completions
+  DEVICE_AFTERMARKET_CONTRACT_ADDRESS: "0x5D666F215a85B87Cb042D59662A7ecd2C8Cc44e6"
```
elffjs, line 79: What's this address? I think on dev it should be `0xcf9af64522162da85164a714c23a7705e6e466b3`, and `0x9c94c395cbcbde662235e0a9d3bb87ad708561ba` on prod.

internal/config/settings.go
elffjs: Format this Go file, I meant.

elffjs (changes requested):
If we fix the values I'm good. I will make a follow-up for tests.

elffjs:
Okay, I did create a slight problem for you: in the old privilege test we have to specify the chain id. My bad.

elffjs:
I'll take care of the print statement. Follow-up for testing: https://linear.app/dimo/issue/PLA-1241/test-aftermarket-listener

PR #50, opened 2023-04-07
[PLA-1268] Change decoding to GRPC call

Asfiroth:
# Proposed Changes
- call to decode vin grpc


internal/services/device_definitions_service.go
```diff
@@ -406,17 +410,30 @@ func (d *deviceDefinitionService) PullDrivlyData(ctx context.Context, userDevice
...
+
+	defer conn.Close()
+
+	_, err = client.DecodeVin(ctx, &ddgrpc.DecodeVinRequest{Vin: vin})
+
+	if err != nil {
+		return ErrorDataPullStatus, errors.Wrap(err, "error pulling vin info from drivly")
+	}
+
 	existingVINData, err := models.ExternalVinData(
```
JamesReate, line 427: i think we may not need this anymore, since we never want to pull VIN data anymore from here, since it will all be handled in the vin decode in device-definitions-api. Could probably also remove the forceSetVinInfo since that metadata is handled also with the decode vin

JamesReate (changes requested):

internal/controllers/helpers/handlers.go
```diff
@@ -149,3 +151,60 @@ func ErrorHandler(c *fiber.Ctx, err error, logger zerolog.Logger, isProduction b
 		"message": err.Error(),
 	})
 }
+
+type DeviceAttributeType string
+
+const (
+	MPG                 DeviceAttributeType = "mpg"
```
JamesReate, line 158:
```suggestion
	Mpg                 DeviceAttributeType = "mpg"
```
just for consistency, or change the other one to be `MPGHighway`

JamesReate (approved):
minor tweak but looks good

PR #56, opened 2023-04-13
PLA-1252-user-device-ownership-middleware

Allyson-English:
# Proposed Changes
check that the authenticated user owns the userDeviceID specified in the path, or has a confirmed Ethereum address that owns the corresponding NFT


internal/controllers/user_devices_controller.go
```diff
@@ -133,6 +133,60 @@ func NewUserDevicesController(settings *config.Settings, dbs func() *db.ReaderWr
 	}
 }
 
+// DeviceOwnershipMiddleWare check that authenticated user owns the userDeviceID in the path or has a confirmed Eth address that owns the corresponding NFT
+func (udc *UserDevicesController) DeviceOwnershipMiddleWare(c *fiber.Ctx) error {
```
elffjs, line 137: Could we divorce this from the controller? It needs a database connection and a users client, anything else?

internal/controllers/user_devices_controller.go
```diff
@@ -133,6 +133,60 @@ func NewUserDevicesController(settings *config.Settings, dbs func() *db.ReaderWr
...
+	if err != nil && !errors.Is(err, sql.ErrNoRows) {
+		udc.log.Err(err).Msg("error while checking if authenticated user owns device or corresponding nft")
+		return err
+	}
+
+	if userDevice != nil {
+		return c.Next()
+	}
+
+	conn, err := grpc.Dial(udc.Settings.UsersAPIGRPCAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
elffjs, line 154: I'd like to avoid dialing every time, can we make this a dependency? I have a branch where `usersClient` is a property on `udc` but I'd like to get away from this mega-controller.

internal/controllers/user_devices_controller.go
```diff
@@ -133,6 +133,60 @@ func NewUserDevicesController(settings *config.Settings, dbs func() *db.ReaderWr
...
 
+// DeviceOwnershipMiddleWare check that authenticated user owns the userDeviceID in the path or has a confirmed Eth address that owns the corresponding NFT
+func (udc *UserDevicesController) DeviceOwnershipMiddleWare(c *fiber.Ctx) error {
+	udi := c.Params("userDeviceID")
+	userID := helpers.GetUserID(c)
+
+	userDevice, err := models.UserDevices(
+		models.UserDeviceWhere.ID.EQ(udi),
+		models.UserDeviceWhere.UserID.EQ(userID),
+	).One(c.Context(), udc.DBS().Reader)
```
elffjs, line 144: Note that you only care about existence, so you could change to `.Exists`.

cmd/devices-api/api.go
```diff
@@ -178,39 +181,39 @@ func startWebAPI(logger zerolog.Logger, settings *config.Settings, pdb db.Store,
...
-	v1Auth.Patch("/user/devices/:userDeviceID/name", userDeviceController.UpdateName)
-	v1Auth.Patch("/user/devices/:userDeviceID/country-code", userDeviceController.UpdateCountryCode)
-	v1Auth.Patch("/user/devices/:userDeviceID/image", userDeviceController.UpdateImage)
-	v1Auth.Get("/user/devices/:userDeviceID/valuations", userDeviceController.GetValuations)
-	v1Auth.Get("/user/devices/:userDeviceID/offers", userDeviceController.GetOffers)
-	v1Auth.Get("/user/devices/:userDeviceID/range", userDeviceController.GetRange)
-	v1Auth.Get("/user/devices/:userDeviceID/status", userDeviceController.GetUserDeviceStatus)
-	v1Auth.Post("/user/devices/:userDeviceID/error-codes", userDeviceController.QueryDeviceErrorCodes)
-	v1Auth.Get("/user/devices/:userDeviceID/error-codes", userDeviceController.GetUserDeviceErrorCodeQueries)
+	udOwner.Delete("/", userDeviceController.DeleteUserDevice)
```
elffjs, line 184: Love it.

internal/controllers/user_devices_controller.go
```diff
@@ -133,6 +133,60 @@ func NewUserDevicesController(settings *config.Settings, dbs func() *db.ReaderWr
...
+		udc.log.Err(err).Msg("error while checking if authenticated user owns corresponding nft")
+		return err
+	}
+
+	l := udc.log.With().
+		Timestamp().
+		Str("user_device_id", udi).Str("user_id", userID).
+		Logger()
+
+	c.SetUserContext(l.WithContext(c.Context()))
```
elffjs, line 186: Maybe we use `c.Locals`? Can we try storing `&l`?

Also, we would still like to store this if we hit line 151.

internal/controllers/user_devices_controller.go
```diff
@@ -133,6 +133,60 @@ func NewUserDevicesController(settings *config.Settings, dbs func() *db.ReaderWr
...
+		models.VehicleNFTWhere.UserDeviceID.EQ(null.StringFrom(udi)),
+	).All(c.Context(), udc.DBS().Reader)
+	if err != nil {
+		udc.log.Err(err).Msg("error while checking if authenticated user owns corresponding nft")
+		return err
+	}
+
+	l := udc.log.With().
+		Timestamp().
+		Str("user_device_id", udi).Str("user_id", userID).
```
elffjs, line 183:
```suggestion
		Str("userDeviceId", udi).Str("userId", userID).
```

internal/controllers/user_devices_controller.go
```diff
@@ -133,6 +133,60 @@ func NewUserDevicesController(settings *config.Settings, dbs func() *db.ReaderWr
...
+		models.VehicleNFTWhere.OwnerAddress.EQ(null.BytesFrom(common.FromHex(*user.EthereumAddress))),
+		models.VehicleNFTWhere.UserDeviceID.EQ(null.StringFrom(udi)),
+	).All(c.Context(), udc.DBS().Reader)
+	if err != nil {
+		udc.log.Err(err).Msg("error while checking if authenticated user owns corresponding nft")
+		return err
+	}
+
+	l := udc.log.With().
+		Timestamp().
```
elffjs, line 182: Already has it, no worries there.

internal/controllers/user_devices_controller.go
```diff
@@ -133,6 +133,60 @@ func NewUserDevicesController(settings *config.Settings, dbs func() *db.ReaderWr
...
+	}
+
+	if user.EthereumAddress == nil {
+		return fiber.NewError(fiber.StatusConflict, "User does not have an Ethereum address and does not own device associated with userDeviceID.")
+	}
+
+	_, err = models.VehicleNFTS(
+		models.VehicleNFTWhere.OwnerAddress.EQ(null.BytesFrom(common.FromHex(*user.EthereumAddress))),
+		models.VehicleNFTWhere.UserDeviceID.EQ(null.StringFrom(udi)),
+	).All(c.Context(), udc.DBS().Reader)
```
elffjs, line 175: Maybe just `.Exists` here too?

I'm not sure this is protecting access as it is—`.All` is perfectly happy to return an empty list.

internal/controllers/user_devices_controller.go
```diff
@@ -133,6 +133,60 @@ func NewUserDevicesController(settings *config.Settings, dbs func() *db.ReaderWr
...
+
+	usersClient := pb.NewUserServiceClient(conn)
+	user, err := usersClient.GetUser(c.Context(), &pb.GetUserRequest{Id: userID})
+	if err != nil {
+		udc.log.Err(err).Msg("Failed to retrieve user information.")
+		return opaqueInternalError
+	}
+
+	if user.EthereumAddress == nil {
+		return fiber.NewError(fiber.StatusConflict, "User does not have an Ethereum address and does not own device associated with userDeviceID.")
```
elffjs, line 169: Maybe this is paranoid, but could we do 404 here too?

internal/controllers/user_devices_controller.go
```diff
@@ -133,6 +133,60 @@ func NewUserDevicesController(settings *config.Settings, dbs func() *db.ReaderWr
...
+	udi := c.Params("userDeviceID")
+	userID := helpers.GetUserID(c)
+
+	userDevice, err := models.UserDevices(
+		models.UserDeviceWhere.ID.EQ(udi),
+		models.UserDeviceWhere.UserID.EQ(userID),
+	).One(c.Context(), udc.DBS().Reader)
+	if err != nil && !errors.Is(err, sql.ErrNoRows) {
+		udc.log.Err(err).Msg("error while checking if authenticated user owns device or corresponding nft")
+		return err
```
elffjs, line 147: This will get logged, so you could think about removing the line above if it's not adding much. Could make this comment in a few places.

internal/controllers/user_devices_controller.go
elffjs: One could consider trying to chain or combine these queries, but I think the strategy here is fine.

elffjs (changes requested):
Will look at the tests soon. We're on the right track.

internal/controllers/user_devices_controller.go
Allyson-English: added `NewMiddleware` to helpers, which accepts the database connection and users client as arguments-- this means that we're now creating the users api client inside startWebAPI so it can be passed in. Is there a reason we weren't doing this before? (I noticed that GetSharedDevices, GetMintDevice, PostMintDevice etc are all establishing a connection separately)

internal/controllers/middleware.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/DIMO-Network/devices-api/models"
+	pb "github.com/DIMO-Network/shared/api/users"
+	"github.com/DIMO-Network/shared/db"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/gofiber/fiber/v2"
+	"github.com/rs/zerolog"
+	"github.com/volatiletech/null/v8"
+)
+
+type middleware struct {
```
elffjs, line 15: Can the name be more specific?

internal/controllers/middleware.go
```diff
@@ -0,0 +1,76 @@
...
+	DBS         func() *db.ReaderWriter
+	UsersClient pb.UserServiceClient
+	log         *zerolog.Logger
+	Settings    *config.Settings
+}
+
+// NewMiddleware if user has valid JWT, confirm that they own the userDeviceID in path or have a confirmed eth address that owns the corresponding nft
+func NewMiddleware(settings *config.Settings, dbs func() *db.ReaderWriter, usersClient pb.UserServiceClient, logger *zerolog.Logger) *middleware {
```
elffjs, line 23: The linter is right! This is odd.

internal/controllers/middleware.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+}
+
+// DeviceOwnershipMiddleware check that authenticated user owns the userDeviceID in the path or has a confirmed Eth address that owns the corresponding NFT
+func (m *middleware) DeviceOwnershipMiddleware(c *fiber.Ctx) error {
+	udi := c.Params("userDeviceID")
+	userID := helpers.GetUserID(c)
+
+	c.Locals("userDeviceId", udi)
+	c.Locals("userId", userID)
```
elffjs, line 38: What about the logger?

internal/controllers/middleware.go
```diff
@@ -0,0 +1,76 @@
...
+		return c.Next()
+	}
+
+	user, err := m.UsersClient.GetUser(c.Context(), &pb.GetUserRequest{Id: userID})
+	if err != nil {
+		m.log.Err(err).Msg("Failed to retrieve user information.")
+		return fiber.NewError(fiber.StatusNotFound, err.Error())
+	}
+
+	if *user.EthereumAddress == "0x0000000000000000000000000000000000000000" {
```
elffjs, line 59: I don't think this is a failure mode we have. The choices are `nil` or `*user.EthereumAddress` is a real hex address.

I regret this API, maybe we'll fix it someday.

internal/controllers/middleware.go
```diff
@@ -0,0 +1,76 @@
...
+		return fiber.NewError(fiber.StatusNotFound, "User does not have an Ethereum address and does not own device associated with userDeviceID.")
+	}
+
+	nft, err := models.VehicleNFTS(
+		models.VehicleNFTWhere.OwnerAddress.EQ(null.BytesFrom(common.FromHex(*user.EthereumAddress))),
+		models.VehicleNFTWhere.UserDeviceID.EQ(null.StringFrom(udi)),
```
elffjs, line 65: This is an interesting idea. I think we're actually not losing any specificity in our responses, which is nice.

internal/controllers/middleware_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+	"github.com/DIMO-Network/devices-api/internal/test"
+	"github.com/gofiber/fiber/v2"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+const genericUserID = "genericUserID"
+
+func TestMiddleware_DeviceDoesNotExist(t *testing.T) {
```
elffjs, line 18: Can we table test these? They're kinda repetitive.

internal/test/helpers.go
```diff
@@ -214,6 +217,29 @@ func SetupCreateUserDevice(t *testing.T, testUserID string, ddID string, metadat
 	return ud
 }
 
+func SetupCreateUserDeviceForMiddleware(t *testing.T, userDeviceID, userID string, ddID string, metadata *[]byte, vin string, pdb db.Store) models.UserDevice {
```
elffjs, line 220: I'm surprised that we can't use something off the shelf. What does this offer beyond `SetupCreateUserDevice`? It's not clear to me.

internal/controllers/user_integrations_controller_test.go
```diff
@@ -101,12 +101,15 @@ func (s *UserIntegrationsControllerTestSuite) SetupSuite() {
 		s.autoPiIngest, s.deviceDefinitionRegistrar, s.autoPiTaskService, nil, nil, s.drivlyTaskSvc, nil, s.redisClient, nil, nil)
 
 	app := test.SetupAppFiber(*logger)
+
 	app.Post("/user/devices/:userDeviceID/integrations/:integrationID", test.AuthInjectorTestHandler(testUserID), c.RegisterDeviceIntegration)
-	app.Post("/user2/devices/:userDeviceID/integrations/:integrationID", test.AuthInjectorTestHandler(testUser2), c.RegisterDeviceIntegration)
-	app.Get("/integrations", c.GetIntegrations)
 	app.Post("/user/devices/:userDeviceID/autopi/command", test.AuthInjectorTestHandler(testUserID), c.SendAutoPiCommand)
 	app.Get("/user/devices/:userDeviceID/autopi/command/:jobID", test.AuthInjectorTestHandler(testUserID), c.GetAutoPiCommandStatus)
+
```
elffjs, line 108: Why'd we move these?

internal/test/helpers.go
```diff
@@ -274,6 +333,42 @@ func SetupCreateUserDeviceAPIIntegration(t *testing.T, autoPiUnitID, externalID,
...
+	ID      string
+	Address common.Address
+}
+
+func (d *FakeUserClient) GetUser(ctx context.Context, in *pb.GetUserRequest, opts ...grpc.CallOption) (*pb.User, error) {
+	for _, user := range d.Response {
+		if user.ID == in.Id {
+			addr := user.Address.Hex()
+
+			if addr == "" {
```
elffjs, line 354: This seems unlikely.

internal/test/helpers.go
```diff
@@ -274,6 +333,42 @@ func SetupCreateUserDeviceAPIIntegration(t *testing.T, autoPiUnitID, externalID,
 	return udapiInt
 }
 
+var MkAddr = func(i int) common.Address {
+	return common.BigToAddress(big.NewInt(int64(i)))
+}
+
+type FakeUserClient struct {
+	Response []UserClientResp
```
elffjs, line 341: I would do like `map[string]*pb.User` and work from there.

cmd/devices-api/api.go
```diff
@@ -82,6 +82,12 @@ func startWebAPI(logger zerolog.Logger, settings *config.Settings, pdb db.Store,
 	autoPi := autopi.NewIntegration(pdb.DBS, ddSvc, autoPiSvc, autoPiTaskService, autoPiIngest, eventService, deviceDefinitionRegistrar, hardwareTemplateService, &logger)
 	openAI := services.NewOpenAI(&logger, *settings)
 
+	conn, err := grpc.Dial(settings.UsersAPIGRPCAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
elffjs, line 85: If you merge in latest—there should be a client readily available.

internal/controllers/middleware.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	if userDevice {
+		return c.Next()
+	}
+
+	user, err := m.UsersClient.GetUser(c.Context(), &pb.GetUserRequest{Id: userID})
+	if err != nil {
+		m.log.Err(err).Msg("Failed to retrieve user information.")
+		return fiber.NewError(fiber.StatusNotFound, err.Error())
```
elffjs, line 56: I think this is an internal error, personally.

internal/controllers/user_integrations_controller_test.go
Allyson-English: I initially moved these because I was trying to incorporate the middleware tests into these and I wanted to visually distinguish what was being grouped (and tested with middleware) the ones I moved all had `/user/devices/:userDeviceID` and were injecting `testUserID` with `AuthInjectorTestHandler`

internal/test/helpers.go
Allyson-English: `SetupCreateUserDevice` generates the user id using `ksuid.New()` but I wanted to be able to pass it in

elffjs:
Monday we start exercising this.

PR #62, opened 2023-04-17
PLA-1261-listen-aftermarket-beneficiary

Allyson-English:
# Proposed Changes

For each aftermarket device, we store a beneficiary address, which may be blank
When a BeneficiarySet event for an aftermarket device comes in, we:
     - Store it if it is non-zero
     - Clear the beneficiary if the provided address is zero
When a Transfer for an aftermarket device comes in, we clear the beneficiary field


internal/services/contracts_events_consumer.go
```diff
@@ -34,9 +34,11 @@ type ContractsEventsConsumer struct {
...
 
 const (
-	PrivilegeSet                EventName = "PrivilegeSet"
-	AftermarketDeviceNodeMinted EventName = "AftermarketDeviceNodeMinted"
-	Transfer                    EventName = "Transfer"
+	PrivilegeSet                 EventName = "PrivilegeSet"
+	AftermarketDeviceNodeMinted  EventName = "AftermarketDeviceNodeMinted"
+	Transfer                     EventName = "Transfer"
+	BeneficiarySet               EventName = "BeneficiarySet"
+	AftermarketDeviceTransferred EventName = "AftermarketDeviceTransferred"
```
elffjs, line 41: The transfer event for aftermarket devices is the standard one, we're already handling it as `Transfer` up above. So I think to handle beneficiary cleanup, we might just want to modify `handleAfterMarketTransferEvent`.

internal/services/contracts_events_consumer.go
```diff
@@ -119,7 +121,12 @@ func (c *ContractsEventsConsumer) processEvent(event *shared.CloudEvent[json.Raw
 			c.log.Info().Str("event", data.EventName).Msg("Event received")
 			return c.setMintedAfterMarketDevice(&data)
 		}
-		fallthrough // TODO(elffjs): Danger!
+	case BeneficiarySet.String():
+		c.log.Info().Str("event", data.EventName).Msg("Event received")
```
elffjs, line 125: We should make sure that this is coming from the aftermarket contract.

internal/services/contracts_events_consumer.go
```diff
@@ -240,3 +247,55 @@ func (c *ContractsEventsConsumer) setMintedAfterMarketDevice(e *ContractEventDat
 
 	return nil
 }
+
+func (c *ContractsEventsConsumer) beneficiarySet(e *ContractEventData) error {
+	var args contracts.RegistryBeneficiarySet
+	err := json.Unmarshal(e.Arguments, &args)
+	if err != nil {
+		return err
+	}
```
elffjs, line 256:
```suggestion
	if err := json.Unmarshal(e.Arguments, &args); err != nil {
		return err
	}
```

internal/services/contracts_events_consumer.go
```diff
@@ -240,3 +247,55 @@ func (c *ContractsEventsConsumer) setMintedAfterMarketDevice(e *ContractEventDat
...
+
+	c.log.Info().Int64("nodeID", args.NodeId.Int64()).Msgf("Aftermarket beneficiary set: %s.", args.Beneficiary)
+
+	device, err := models.AutopiUnits(
+		models.AutopiUnitWhere.AutopiUnitID.EQ(args.NodeId.String()),
+	).One(context.Background(), c.db.DBS().Reader)
+
+	cols := models.AutopiUnitColumns
+
+	device.Beneficiary = null.BytesFrom(args.Beneficiary[:])
```
elffjs, line 266: I think we do want to handle the case of `args.Beneficiary == common.Address{}` differently, and in that case set this field to null.

internal/services/contracts_events_consumer.go
```diff
@@ -240,3 +247,55 @@ func (c *ContractsEventsConsumer) setMintedAfterMarketDevice(e *ContractEventDat
...
+
+	err = device.Upsert(context.Background(), c.db.DBS().Writer, true, []string{cols.Beneficiary}, boil.Infer(), boil.Infer())
```
elffjs, line 268: I would just do an update here, of `beneficiary` and `updated_at`.

migrations/20230417142942_add_beneficiary_addr_to_autopi_tbl.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+-- +goose StatementBegin
+SET search_path = devices_api, public;
+
+ALTER TABLE autopi_units
+    ADD COLUMN beneficiary bytea;
```
elffjs, line 6: Constraint?

elffjs (changes requested):
We will get the corrected ABI from Lorran, and do some tests.

elffjs:
Notes from conversation:

* Do we ignore other contracts?
* Can we go from null to explicitly set beneficiary?
* Can we go from one explicitly set beneficiary to another?
* Can we go from beneficiary to explicitly cleared beneficiary?
* Can we go from beneficiary to cleared after transfer?

internal/contracts/registry.go
```diff
@@ -55,7 +54,7 @@ type DevAdminIdManufacturerName struct {
...
...":\"address\"},{\"indexed\":true,\"internalType\":\"uint256\",\"name\":\"nodeId\",\"type\":\"uint256\"},{\"indexed\":true,\"internalType\":\"address\",\"name\":\"beneficiary\",\"type\":\"address\"}],\"name\":\"BeneficiarySet\",\"type\":\"event\"},{\"inputs\":[{\"internalType\":\"address\",\"name\":\"idProxyAddress\",\"type\":\"address\"},{\"internalType\":\"uint256\",\"name\":\"nodeId\",\"type\":\"uint256\"}],\"name\":\"getBeneficiary\",\"outputs\":[{\"internalType\":\"address\",\"name\":\"beneficiary\",\"type\":\"address\"}],\"stateMutability\":\"view\",\"type\":\"function\"},{\"inputs\":[{\"internalType\":\"address\",\"name\":\"idProxyAddress\",\"type\":\"address\"},{\"internalType\":\"uint256\",\"name\":\"sourceNode\",\"type\":\"uint256\"}],\"name\":\"getLink\",\"outputs\":[{\"internalType\":\"uint256\",\"name\":\"targetNode\",\"type\":\"uint256\"}],\"stateMutability\":\"view\",\"type\":\"function\"},{\"inputs\":[{\"internalType\":\"uint256\",\"name\":\"nodeId\",\"type\":\"uint256\"},{\"internalType\":\"address\",\"name\":\"beneficiary\",\"type\":\"address\"}],\"name\":\"setAftermarketDeviceBeneficiary\",\"outputs\":[],\"stateMutability\":\"nonpayable\",\"type\":\"function\"}]",
```
elffjs, line 57: I think we still have the bum ABI version with the two `AftermarketDeviceUnpaired` events. Could you pull again?

internal/services/contracts_events_consumer.go
```diff
@@ -178,11 +184,11 @@ func (c *ContractsEventsConsumer) handleAfterMarketTransferEvent(e *ContractEven
 
 	apUnit.UserID = null.String{}
 	apUnit.OwnerAddress = null.BytesFrom(args.To.Bytes())
+	apUnit.Beneficiary = null.Bytes{}
 
 	cols := models.AutopiUnitColumns
 
-	_, err = apUnit.Update(ctx, c.db.DBS().Writer, boil.Whitelist(cols.UserID, cols.OwnerAddress))
-	if err != nil {
+	if _, err = apUnit.Update(ctx, c.db.DBS().Writer, boil.Whitelist(cols.UserID, cols.OwnerAddress, cols.Beneficiary)); err != nil {
```
elffjs, line 191: Could you do us a small favor and include `updated_at` here too? I know this was a problem before you got her ebut

internal/services/contracts_events_consumer_test.go
```diff
@@ -339,6 +344,122 @@ func Test_Ignore_Transfer_Unit_Not_Found(t *testing.T) {
...
+			},
+			AutopiUnitTable: models.AutopiUnit{
+				OwnerAddress: null.BytesFrom(common.BigToAddress(big.NewInt(1)).Bytes()),
+				TokenID:      types.NewNullDecimal(new(decimal.Big).SetBigMantScale(big.NewInt(3), 0)),
+				Beneficiary:  null.BytesFrom(common.BigToAddress(big.NewInt(2)).Bytes()),
+			},
+			ExpectedBeneficiaryResult: null.BytesFrom(common.BigToAddress(big.NewInt(3)).Bytes()),
+		},
+		{
+			Name:    "Go from beneficiary to explicitly cleared beneficiary",
```
elffjs, line 409: I think the one thing we should probably check still is: I have a unit, the beneficiary is set, then I get a transfer event—do I clear this thing correctly?

elffjs (approved):

models/token_transfers.go
elffjs, line 1: What is this token transfers thing?

elffjs:
Updated the ABI.

models/token_transfers.go
elffjs, line 1: I'm going to assume this is in here by mistake.

internal/services/contracts_events_consumer.go
elffjs: I'll just do it.

PR #65, opened 2023-04-19
PLA-1316-devices-owned-by-addr

Allyson-English:
# Proposed Changes

- adjust `/user/devices/me` endpoint to include vehicles that are owned by the user's eth address, taking care to not double-show vehicles

### DIFF
--- a/internal/controllers/user_devices_controller.go
+++ b/internal/controllers/user_devices_controller.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/DIMO-Network/shared/redis"
+	"golang.org/x/exp/slices"
 
 	ddgrpc "github.com/DIMO-Network/device-definitions-api/pkg/grpc"
 	"github.com/DIMO-Network/devices-api/internal/config"
@@ -164,7 +165,13 @@ func (udc *UserDevicesController) dbDevicesToDisplay(ctx context.Context, device
 		return nil, helpers.GrpcErrorToFiber(err, "failed to get integrations")
 	}
 
+	seen := make([]string, 0)
 	for _, d := range devices {
+		if slices.Contains(seen, d.ID) {
+			continue
+		}
+		seen = append(seen, d.ID)
+
 		deviceDefinition, err := filterDeviceDefinition(d.DeviceDefinitionID, deviceDefinitionResponse)
 		if err != nil {
 			return nil, err
@@ -277,7 +284,7 @@ func (udc *UserDevicesController) GetUserDevices(c *fiber.Ctx) error {
 	// todo grpc call out to grpc service endpoint in the deviceDefinitionsService udc.deviceDefSvc.GetDeviceDefinitionsByIDs(c.Context(), []string{ "todo"} )
 
 	userID := helpers.GetUserID(c)
-	devices, err := models.UserDevices(
+	userOwnedDevices, err := models.UserDevices(
 		models.UserDeviceWhere.UserID.EQ(userID),
 		qm.Load(models.UserDeviceRels.UserDeviceAPIIntegrations),
 		qm.Load(qm.Rels(models.UserDeviceRels.VehicleNFT, models.VehicleNFTRels.MintRequest)),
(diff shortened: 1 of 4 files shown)


internal/controllers/user_devices_controller.go
```diff
@@ -164,7 +165,13 @@ func (udc *UserDevicesController) dbDevicesToDisplay(ctx context.Context, device
 		return nil, helpers.GrpcErrorToFiber(err, "failed to get integrations")
 	}
 
+	seen := make([]string, 0)
```
elffjs, line 168: What's this `seen` thing? Are we using this?

internal/controllers/user_devices_controller.go
```diff
@@ -277,8 +284,29 @@ func (udc *UserDevicesController) GetUserDevices(c *fiber.Ctx) error {
 	// todo grpc call out to grpc service endpoint in the deviceDefinitionsService udc.deviceDefSvc.GetDeviceDefinitionsByIDs(c.Context(), []string{ "todo"} )
 
 	userID := helpers.GetUserID(c)
+	user, err := udc.usersClient.GetUser(c.Context(), &pb.GetUserRequest{Id: userID})
+	if err != nil {
+		return helpers.ErrorResponseHandler(c, err, fiber.StatusInternalServerError)
+	}
+
+	addr := common.Hex2Bytes(*user.EthereumAddress)
+	// TODO(AE): could reduce db calls with a nested select in UserDevices
```
elffjs, line 293: I would like to see a left outer join with `vehicle_nfts`, to see how well that performs.

internal/controllers/user_devices_controller_test.go
```diff
@@ -400,8 +404,13 @@ func (s *UserDevicesControllerTestSuite) TestGetMyUserDevices() {
 	)
 	_ = test.SetupCreateAutoPiUnit(s.T(), testUserID, unitID, func(s string) *string { return &s }(deviceID), s.pdb)
 	_ = test.SetupCreateUserDeviceAPIIntegration(s.T(), unitID, deviceID, ud.ID, integration.Id, s.pdb)
+
+	addr := "67B94473D81D0cd00849D563C94d0432Ac988B49"
```
elffjs, line 408: I think the goal of these test additions is a bit unclear. Are we just trying to prevent an NPE?

internal/controllers/user_devices_controller.go
```diff
@@ -277,8 +284,29 @@ func (udc *UserDevicesController) GetUserDevices(c *fiber.Ctx) error {
 	// todo grpc call out to grpc service endpoint in the deviceDefinitionsService udc.deviceDefSvc.GetDeviceDefinitionsByIDs(c.Context(), []string{ "todo"} )
 
 	userID := helpers.GetUserID(c)
+	user, err := udc.usersClient.GetUser(c.Context(), &pb.GetUserRequest{Id: userID})
+	if err != nil {
+		return helpers.ErrorResponseHandler(c, err, fiber.StatusInternalServerError)
+	}
+
+	addr := common.Hex2Bytes(*user.EthereumAddress)
```
elffjs, line 292: Does this have to be non-nil? Isn't this a bit dangerous?

elffjs (changes requested):
See comments!

internal/controllers/user_devices_controller.go
Allyson-English: this is how I'm  making sure we don't include duplicates in the final response:

```
	seen := make([]string, 0)
	for _, d := range devices {
		if slices.Contains(seen, d.ID) {
			continue
		}
		seen = append(seen, d.ID)
```

internal/controllers/user_devices_controller_test.go
```diff
@@ -391,15 +395,69 @@ func (s *UserDevicesControllerTestSuite) TestPostInvalidDefinitionID() {
...
+	assert.Equal(s.T(), fiber.StatusOK, response.StatusCode)
+	result := gjson.Get(string(body), "userDevices.#.id")
+	assert.Len(s.T(), result.Array(), 2)
+	for _, id := range result.Array() {
+		assert.True(s.T(), id.Exists(), "expected to find the ID")
+	}
+
+	assert.Equal(s.T(), integration.Id, gjson.GetBytes(body, "userDevices.1.integrations.0.integrationId").String())
+	assert.Equal(s.T(), deviceID, gjson.GetBytes(body, "userDevices.1.integrations.0.externalId").String())
+	assert.Equal(s.T(), integration.Vendor, gjson.GetBytes(body, "userDevices.1.integrations.0.integrationVendor").String())
```
elffjs, line 436: I think this test is probably doing too much at this point. But let's move on for now! This is good! Testing is good.

elffjs (approved):

PR #68, opened 2023-04-20
Remove wrong messaging from contracts events listener

0xdev22:
# Proposed Changes

Remove returned error for non Aftermarket events and change error message for contract processing

### DIFF
--- a/internal/services/contracts_events_consumer.go
+++ b/internal/services/contracts_events_consumer.go
@@ -143,7 +143,7 @@ func (c *ContractsEventsConsumer) routeTransferEvent(e *ContractEventData) error
 		c.log.Debug().Str("event", e.EventName).Interface("fullEventData", e).Msg("Handler not provided for contract")
 	}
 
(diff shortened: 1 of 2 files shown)


internal/services/contracts_events_consumer.go
```diff
@@ -143,7 +143,7 @@ func (c *ContractsEventsConsumer) routeTransferEvent(e *ContractEventData) error
 		c.log.Debug().Str("event", e.EventName).Interface("fullEventData", e).Msg("Handler not provided for contract")
 	}
 
-	return errors.New("Handler not provided for contract")
+	return nil
```
elffjs, line 146: I sort of wanted to avoid "ignoring mint event" and "device has not been claimed yet". Those happen a lot.

elffjs (changes requested):
Asking for a little more, as usual. Sorry!

elffjs (approved):

PR #71, opened 2023-04-20
Pla 1342 grpc list user nft devices

0xdev22:
# Proposed Changes


pkg/grpc/aftermarket_devices.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.28.1
```
elffjs, line 3: Might want to upgrade local protoc.

internal/api/user_devices.go
```diff
@@ -111,6 +112,26 @@ func (s *userDeviceService) ListUserDevicesForUser(ctx context.Context, req *pb.
 	return &pb.ListUserDevicesForUserResponse{UserDevices: out}, nil
 }
 
+func (s *userDeviceService) ListDevicesForWalletAddress(ctx context.Context, req *pb.ListDevicesForWalletAddressRequest) (*pb.ListDevicesForWalletAddressResponse, error) {
+	vNft, err := models.VehicleNFTS(
+		models.VehicleNFTWhere.OwnerAddress.EQ(null.BytesFrom(common.FromHex(req.UserAddress))),
+		qm.Load(models.VehicleNFTRels.UserDevice),
```
elffjs, line 118: We want integration information, as on line 99. I think we are missing that here.

internal/api/user_devices.go
```diff
@@ -111,6 +112,26 @@ func (s *userDeviceService) ListUserDevicesForUser(ctx context.Context, req *pb.
...
+
+	if err != nil {
+		s.logger.Err(err).Str("eth address", req.UserAddress).Msg("Database failure retrieving user's devices.")
+		return nil, status.Error(codes.Internal, "Internal error.")
+	}
+
+	out := make([]*pb.UserDevice, len(vNft))
+
+	for i := 0; i < len(vNft); i++ {
+		out[i] = s.deviceModelToAPI(vNft[i].R.UserDevice)
```
elffjs, line 129: Will these guys have `vNft[i].R.UserDevice.R.VehicleNFT` populated? I honestly don't know.

elffjs (changes requested):

internal/api/user_devices.go
elffjs: It looks like the answer is yes!

internal/api/user_devices.go
elffjs: We also want NFT pairing information, as on line 98, we go into the autopi_units table.

elffjs:
Maybe this will help motivate: we'd like to use this here:

https://github.com/DIMO-Network/rewards-api/blob/cc4765e25f45e505b2c886748e4c7fc49957411f/internal/controllers/rewards.go#L75

internal/api/user_devices.go
```diff
@@ -113,13 +114,17 @@ func (s *userDeviceService) ListUserDevicesForUser(ctx context.Context, req *pb.
 }
 
 func (s *userDeviceService) ListDevicesForWalletAddress(ctx context.Context, req *pb.ListDevicesForWalletAddressRequest) (*pb.ListDevicesForWalletAddressResponse, error) {
+	if req.UserAddress == "" {
```
elffjs, line 117: Maybe we should go all the way and validate the address? `common.IsHexAddress`?

internal/api/user_devices.go
```diff
@@ -111,6 +113,30 @@ func (s *userDeviceService) ListUserDevicesForUser(ctx context.Context, req *pb.
 	return &pb.ListUserDevicesForUserResponse{UserDevices: out}, nil
 }
 
+func (s *userDeviceService) ListDevicesForWalletAddress(ctx context.Context, req *pb.ListDevicesForWalletAddressRequest) (*pb.ListDevicesForWalletAddressResponse, error) {
```
elffjs, line 116: I wonder if we should fold this into the old endpoint. We left outer join `user_devices` with `vehicle_nfts` and do `user_devices.user_id = ? OR vehicle_nfts.owner_address = ?`?

internal/api/user_devices.go
```diff
@@ -111,6 +113,30 @@ func (s *userDeviceService) ListUserDevicesForUser(ctx context.Context, req *pb.
...
+	if req.UserAddress == "" {
+		return nil, status.Error(codes.InvalidArgument, "Missing wallet address arguement")
+	}
+	vNft, err := models.VehicleNFTS(
+		models.VehicleNFTWhere.OwnerAddress.EQ(null.BytesFrom(common.FromHex(req.UserAddress))),
+		qm.Load(models.VehicleNFTRels.UserDevice),
+		qm.Load(models.VehicleNFTRels.VehicleTokenAutopiUnit),
+		qm.Load(qm.Rels(models.VehicleNFTRels.UserDevice, models.UserDeviceRels.UserDeviceAPIIntegrations)),
```
elffjs, line 124: One comment: this automatically does the load from line 122.

internal/api/user_devices.go
elffjs: Or maybe a flag on the existing `ListUserDevicesForUser`? I'm not sure.

PR #88, opened 2023-05-01
Clear out privileges when vehicle transfer event is recieved

0xdev22:
# Proposed Changes

### DIFF
--- a/internal/services/contracts_events_consumer.go
+++ b/internal/services/contracts_events_consumer.go
@@ -202,14 +202,12 @@ func (c *ContractsEventsConsumer) handleVehicleTransfer(e *ContractEventData) er
 
 	rowsAff, err := models.NFTPrivileges(
 		models.NFTPrivilegeWhere.TokenID.EQ(types.Decimal(tkID)),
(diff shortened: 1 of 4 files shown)


internal/services/contracts_events_consumer.go
```diff
@@ -200,6 +200,17 @@ func (c *ContractsEventsConsumer) handleVehicleTransfer(e *ContractEventData) er
 
 	defer tx.Rollback() //nolint
 
+	rowsAff, err := models.NFTPrivileges(
+		models.NFTPrivilegeWhere.TokenID.EQ(types.Decimal(tkID)),
+		models.NFTPrivilegeWhere.UserAddress.EQ(args.To.Bytes()),
```
elffjs, line 205: I don't think we want any filter on `user_address`, all privileges must go. The "`to`" is the only one who will have privileges when this is over—of course, he has them implicitly.

elffjs (changes requested):

elffjs (approved):

elffjs:
One request before we leave this ticket alone: Test this out on dev.

* Mint a car (I can show you how to do this)
* Set some privileges
* Transfer the car

PR #93, opened 2023-05-04
PLA-1393-valuation-pct-growth

Allyson-English:
# Proposed Changes

- update valuation percent growth formula to be relative change
- also applied this change to `GetClaimedVehiclesGrowth` which wasn't where we were seeing the issue but I noticed this growth percentage also wasn't using the relative change formula


JamesReate (approved):
lgtm

PR #99, opened 2023-05-10
Endpoint for fetching integration metadata

0xdev22:
# Proposed Changes


elffjs:
Might we mark this as a draft, given that it won’t compile?

cmd/devices-api/api.go
```diff
@@ -193,6 +193,7 @@ func startWebAPI(logger zerolog.Logger, settings *config.Settings, pdb db.Store,
 	v1Auth.Post("/user/devices", userDeviceController.RegisterDeviceForUser)
 
 	v1Auth.Get("/integrations", userDeviceController.GetIntegrations)
+	v1Auth.Get("/integration/:tokenID", userDeviceController.GetIntegrationNFTMetadata)
```
elffjs, line 196: You don’t want to require authentication for this, it’s for public use.

Could we place this next to the other NFT endpoints?

internal/controllers/user_integrations_controller.go
```diff
@@ -220,6 +220,33 @@ func (udc *UserDevicesController) GetIntegrations(c *fiber.Ctx) error {
 	})
 }
 
+// GetIntegrationByTokenID godoc
+// @Description gets an integration using it's tokenID
+// @Tags        integrations
+// @Produce     json
+// @Success     200 {array} ddgrpc.Integration
+// @Security    BearerAuth
```
elffjs, line 228:
```suggestion
```

internal/controllers/user_integrations_controller.go
```diff
@@ -220,6 +220,33 @@ func (udc *UserDevicesController) GetIntegrations(c *fiber.Ctx) error {
 	})
 }
 
+// GetIntegrationByTokenID godoc
+// @Description gets an integration using it's tokenID
+// @Tags        integrations
+// @Produce     json
+// @Success     200 {array} ddgrpc.Integration
```
elffjs, line 227: Is this the type we’re returning?

internal/controllers/user_integrations_controller.go
```diff
@@ -220,6 +220,33 @@ func (udc *UserDevicesController) GetIntegrations(c *fiber.Ctx) error {
 	})
 }
 
+// GetIntegrationByTokenID godoc
+// @Description gets an integration using it's tokenID
```
elffjs, line 224:
```suggestion
// @Description gets an integration using its tokenID
```

internal/controllers/user_integrations_controller.go
```diff
@@ -220,6 +220,33 @@ func (udc *UserDevicesController) GetIntegrations(c *fiber.Ctx) error {
...
+// @Produce     json
+// @Success     200 {array} ddgrpc.Integration
+// @Security    BearerAuth
+// @Router      /integration/:tokenID [get]
+func (udc *UserDevicesController) GetIntegrationNFTMetadata(c *fiber.Ctx) error {
+	tokenID := c.Params("tokenID")
+
+	uTokenID, err := strconv.ParseUint(tokenID, 10, 64)
+	if err != nil {
+		return fiber.NewError(fiber.StatusInternalServerError, "failed to get integration")
```
elffjs, line 235: I think in this case we can say that the token id isn’t a valid number. Probably not an internal error. (If this happens elsewhere in this file then I am sorry.)

elffjs (changes requested):

elffjs (changes requested):
A few comments.

elffjs:
Oh, remember to regen docs.

internal/controllers/user_integrations_controller.go
elffjs: Was this resolved?

internal/controllers/user_integrations_controller.go
elffjs: Going to unresolve and commit. I'm not sure what happened here.

internal/controllers/user_integrations_controller.go
0xdev22: I moved the function to the `nftController`. Its been resolved there

elffjs (approved):

PR #100, opened 2023-05-11
[EXP-427] fix null assignment for total valuation

Asfiroth:
# Proposed Changes
- fix null assignment for total valuation


JamesReate:
gonna merge this since will fix customer facing issue, and reduce unnecessary calls to db. failing tests are from previous commit.

PR #104, opened 2023-05-16
Change response format of error codes endpoint and fix failing tests

0xdev22:
# Proposed Changes


internal/services/openai.go
```diff
@@ -108,32 +114,44 @@ func (o *openAI) askChatGPT(body io.Reader) (*ChatGPTResponse, error) {
...
 
-	return strings.Trim(c.Message.Content, "\n"), nil
+	resp := []ErrorCodesResponse{}
+
+	codesResp := strings.Split(r.Choices[0].Message.Content, "\n")
+	for _, code := range codesResp {
+		cc := strings.Split(code, ":")
+		resp = append(resp, ErrorCodesResponse{
+			Code:        cc[0],
+			Description: cc[1],
```
elffjs, line 152: It seems to me that this always has a space at the start. Just from playing around in the web app.

internal/services/contracts_events_consumer_test.go
```diff
@@ -102,8 +101,6 @@ func TestProcessContractsEventsMessages(t *testing.T) {
 		PrivilegeID: args.privilegeID,
 	}
 
-	log.Println(actual.ExpiresAt, "actual time", expected.ExpiresAt, "expected")
```
elffjs, line 105: 👍

internal/services/openai.go
elffjs, line 152: Just from playing around in the web app: this seems to always have a space at the front.

Note that we are not really testing this code.

internal/services/openai.go
```diff
@@ -108,32 +114,44 @@ func (o *openAI) askChatGPT(body io.Reader) (*ChatGPTResponse, error) {
...
+func (o *openAI) GetErrorCodesDescription(make, model string, errorCodes []string) ([]ErrorCodesResponse, error) {
 	codes := strings.Join(errorCodes, ", ")
+
 	req := fmt.Sprintf(`{
 		"model": "gpt-3.5-turbo",
 		"messages": [
 			{
 				"role": "user", 
-				"content": "Briefly summarize for me, in order, what the following error codes mean for a %s %s. The error codes are %s."}]
+				"content": "A %s %s is returning error codes %s. Respond only with a line for each code, in the format 'code:explanation \n'"}]
```
elffjs, line 125: The newline caught me off guard, but it seems to respond just fine.

internal/services/openai.go
```diff
@@ -108,32 +114,44 @@ func (o *openAI) askChatGPT(body io.Reader) (*ChatGPTResponse, error) {
...
 	if c.FinishReason != "stop" {
 		o.logger.Error().Interface("rawResponse", r).Msg("Unexpected finish_reason from ChatGPT.")
 	}
 
-	return strings.Trim(c.Message.Content, "\n"), nil
+	resp := []ErrorCodesResponse{}
+
+	codesResp := strings.Split(r.Choices[0].Message.Content, "\n")
+	for _, code := range codesResp {
+		cc := strings.Split(code, ":")
```
elffjs, line 149: One concern here is that the description might have a `:` in it, and then you'd lose everything after that.

I believe there is a [`strings.SplitN`](https://pkg.go.dev/strings#SplitN).

elffjs: In the future it would be nice to be able to detect the case in which they send us back something in the wrong format.

migrations/20230516033000_change_error_codes_format.sql
```diff
@@ -0,0 +1,13 @@
+-- +goose Up
+-- +goose StatementBegin
+SET search_path = devices_api, public;
+
+ALTER TABLE devices_api.error_code_queries ADD column codes_query_response jsonb;
```
elffjs, line 5: We have now `query_response` and `codes_query_response`. That's confusing—do you have plans for the older text column?

internal/controllers/device_data_controller.go
```diff
@@ -30,7 +31,7 @@ type QueryDeviceErrorCodesReq struct {
 }
 
 type QueryDeviceErrorCodesResponse struct {
-	Message string `json:"message"`
```
elffjs, line 33: Oh no we gotta regen docs 😁.

internal/controllers/device_data_controller.go
```diff
@@ -30,7 +31,7 @@ type QueryDeviceErrorCodesReq struct {
...
+	Message []services.ErrorCodesResponse `json:"message"`
 }
 
 type GetUserDeviceErrorCodeQueriesResponse struct {
```
elffjs, line 37: Now, I'm worried about these response types. Shouldn't these change? I think this is the most serious concern I have.

elffjs (changes requested):
One major concern.

I will fix the lint error from main, that's my fault.

migrations/20230516033000_change_error_codes_format.sql
0xdev22: I'm leaving it there since its already deployed and the column probably already holds some data

elffjs: Well we’re pre-release, I think it’s okay to run around deleting.

0xdev22: ok

internal/controllers/device_data_controller.go
```diff
@@ -30,17 +31,16 @@ type QueryDeviceErrorCodesReq struct {
...
 	Queries []GetUserDeviceErrorCodeQueriesResponseItem `json:"queries"`
 }
 
 type GetUserDeviceErrorCodeQueriesResponseItem struct {
-	Codes       []string  `json:"errorCodes"`
-	Description string    `json:"description"`
-	RequestedAt time.Time `json:"requestedAt"`
+	ErrorCodesDescription []services.ErrorCodesResponse `json:"errorCodesDescription"`
```
elffjs, line 42: I would do "errorCodeDescriptions` or just `errorCodes`. It feels odd to not have the whole thing be plural.

internal/controllers/device_data_controller.go
```diff
@@ -349,7 +348,13 @@ func (udc *UserDevicesController) QueryDeviceErrorCodes(c *fiber.Ctx) error {
...
 	}
 
-	q := &models.ErrorCodeQuery{ID: ksuid.New().String(), UserDeviceID: udi, ErrorCodes: req.ErrorCodes, QueryResponse: chtResp}
+	chtJSON, err := json.Marshal(chtResp)
+	if err != nil {
+		logger.Err(err).Interface("requestBody", req).Msg("Error occurred fetching description for error codes")
+		return fiber.NewError(fiber.StatusInternalServerError, "Error occurred fetching description for error codes")
+	}
+
+	q := &models.ErrorCodeQuery{ID: ksuid.New().String(), UserDeviceID: udi, ErrorCodes: req.ErrorCodes, CodesQueryResponse: null.JSONFrom(chtJSON)}
```
elffjs, line 357: Do we need both of these database fields?

My guess is we're going to want to spit out code → description into its own table soon enough.

internal/services/openai.go
```diff
@@ -108,32 +114,44 @@ func (o *openAI) askChatGPT(body io.Reader) (*ChatGPTResponse, error) {
...
+		return nil, errors.New("could not fetch description for error codes")
 	}
 
 	c := r.Choices[0]
 	if c.FinishReason != "stop" {
 		o.logger.Error().Interface("rawResponse", r).Msg("Unexpected finish_reason from ChatGPT.")
 	}
 
-	return strings.Trim(c.Message.Content, "\n"), nil
+	codesResp := strings.SplitN(r.Choices[0].Message.Content, "\n", len(errorCodes))
```
elffjs, line 145: Sorry, I think this split was fine. The problem was the split below.

internal/services/openai.go
elffjs: I think the worry is this split here. If you get a document back like
```
XDL1: Car on fire: run for shelter
```
then the code as it is will turn this into
```
{
  "code": "XDL1",
  "description": " Car on fire"
}
```
and drop the second part. A test would pick this up.

You can also ask it to return JSON! I was too lazy to do this and it tends to try to put some intro text.

migrations/20230516033000_change_error_codes_format.sql
elffjs: Eventually I'd like to make this non-nullable, but I understand the hesitancy.

To be clear: I don't think the old column is populated at all on prod.

internal/controllers/nft_controller.go
```diff
@@ -236,7 +236,11 @@ func (nc *NFTController) GetDCNNFTImage(c *fiber.Ctx) error {
 
 	c.Set("Content-Type", "image/svg+xml")
 
-	nc.dcnTmpl.Execute(c, struct{ Name string }{dcn.Name.String})
+	err = nc.dcnTmpl.Execute(c, struct{ Name string }{dcn.Name.String})
```
elffjs, line 239: Ah, thanks, you didn't have to.

internal/controllers/device_data_controller.go
elffjs: How about `errorCodes` just to match the listing response? I don't see a reason for them to differ.

elffjs (changes requested):
I would be happy to merge this as-is, it's in good shape. As long as we're here I think we should treat the `.Split` thing.

internal/services/openai.go
```diff
@@ -108,32 +114,44 @@ func (o *openAI) askChatGPT(body io.Reader) (*ChatGPTResponse, error) {
...
 
-	return strings.Trim(c.Message.Content, "\n"), nil
+	codesResp := strings.Split(r.Choices[0].Message.Content, "\n")
+
+	resp := []ErrorCodesResponse{}
+	for _, code := range codesResp {
+		cc := strings.SplitN(code, ":", 2)
+		resp = append(resp, ErrorCodesResponse{
+			Code:        cc[0],
+			Description: strings.TrimSpace(cc[1]),
```
elffjs, line 152: Slight chance of panicking here. Let's live with it for now.

migrations/20230516033000_change_error_codes_format.sql
elffjs: I'd like this to be non-nullable eventually.

PR #107, opened 2023-05-19
Create endpoint for clearing error codes

0xdev22:
# Proposed Changes


internal/controllers/device_data_controller.go
```diff
@@ -407,6 +408,71 @@ func (udc *UserDevicesController) GetUserDeviceErrorCodeQueries(c *fiber.Ctx) er
...
+// @Success     200 {object} controllers.QueryDeviceErrorCodesResponse
+// @Security    BearerAuth
+// @Router      /user/devices/{userDeviceID}/error-codes/clear [post]
+func (udc *UserDevicesController) ClearUserDeviceErrorCodeQuery(c *fiber.Ctx) error {
+	userID := helpers.GetUserID(c)
+	udi := c.Params("userDeviceID")
+
+	logger := helpers.GetLogger(c, udc.log)
+
+	userDeviceExists, err := models.UserDevices(
```
elffjs, line 423: The middleware will check for ownership, potentially based on anddress, and existence. I would drop this whole thing.

internal/controllers/device_data_controller.go
```diff
@@ -407,6 +408,71 @@ func (udc *UserDevicesController) GetUserDeviceErrorCodeQueries(c *fiber.Ctx) er
...
+		}
+
+		logger.Err(err).Msg("error occurred when fetching error codes for device")
+		return fiber.NewError(fiber.StatusInternalServerError, "error occurred updating device error queries")
+	}
+
+	if !userDeviceExists {
+		return fiber.NewError(fiber.StatusUnauthorized, "error, user does not own device")
+	}
+
```
elffjs, line 439: Are you imagining that they could go into an an historical query and clear that? Physically but I don’t think it’s a, I don’t think it makes sense.

Let’s ask mobile.

internal/controllers/device_data_controller.go
```diff
@@ -407,6 +408,71 @@ func (udc *UserDevicesController) GetUserDeviceErrorCodeQueries(c *fiber.Ctx) er
...
+		req := &ClearUserDeviceErrorCodeQueryReq{}
+		   	if err := c.BodyParser(req); err != nil {
+		   		return fiber.NewError(fiber.StatusBadRequest, "Couldn't parse request.")
+		   	}
+	*/
+
+	errCodeQuery, err := models.ErrorCodeQueries(
+		models.ErrorCodeQueryWhere.UserDeviceID.EQ(udi),
+		models.ErrorCodeQueryWhere.ClearedAt.IsNull(),
```
elffjs, line 451: I would drop this. If the most recent query has been cleared, then I don’t think this endpoint has anything to do..

internal/controllers/device_data_controller_test.go
```diff
@@ -727,3 +728,90 @@ func TestUserDevicesController_GetUserDevicesErrorCodeQueries(t *testing.T) {
 		test.TruncateTables(pdb.DBS().Writer.DB, t)
 	})
 }
+
+func TestUserDevicesController_ClearUserDeviceErrorCodeQuery(t *testing.T) {
+	mockDeps := createMockDependencies(t)
+	defer mockDeps.mockCtrl.Finish()
+
+	ctx := context.Background()
+	pdb, container := test.StartContainerDatabase(ctx, t, migrationsDirRelPath)
```
elffjs, line 737: We’ll have to find a way, at some point, to avoid starting this with every test. These are taking longer and longer to run.

internal/controllers/device_data_controller_test.go
```diff
@@ -727,3 +728,90 @@ func TestUserDevicesController_GetUserDevicesErrorCodeQueries(t *testing.T) {
...
+		errCodeQuery, err := models.ErrorCodeQueries(
+			models.ErrorCodeQueryWhere.ClearedAt.IsNotNull(),
+			qm.OrderBy(models.ErrorCodeQueryColumns.CreatedAt+" DESC"),
+			qm.Limit(1),
+		).One(ctx, pdb.DBS().Reader)
+		assert.NoError(t, err)
+
+		//  currTime.Format(time.RFC3339Nano)
+		assert.JSONEq(t,
+			fmt.Sprintf(`{"errorCodes":%s, "clearedAt":"%v"}`, string(errCodeQuery.CodesQueryResponse.JSON), errCodeQuery.ClearedAt.Time),
```
elffjs, line 810: There is, unfortunately, a difference between the standard Go string and the RFC-3339 format they use for JSON. Easy fix.

elffjs (changes requested):
The overall shape is good. I think some of the old tests are failing because the response has changed.

elffjs:
Also, I can generate the docs. I don’t know why that program is so finicky.

internal/controllers/device_data_controller.go
0xdev22: So on the UI i can see a list and a clear button, which tells me they can choose to select what to clear. What bothers me is if they can choose what to clear, then clearing based on the last created error code wouldn't make sense in that case

internal/controllers/device_data_controller_test.go
0xdev22: yeah, we need something like a beforeAll or maybe a way to put everything in a global variable and run it from there

elffjs:
Merge conflicts ahoy

internal/controllers/device_data_controller_test.go
```diff
@@ -351,9 +358,12 @@ func TestUserDevicesController_QueryDeviceErrorCodes(t *testing.T) {
 		response, _ := app.Test(request)
 		body, _ := io.ReadAll(response.Body)
 
+		chtJSON, err := json.Marshal(chatGptResp)
+		assert.NoError(t, err)
+
 		assert.Equal(t, fiber.StatusOK, response.StatusCode)
 		assert.Equal(t,
-			fmt.Sprintf(`{"message":"%s"}`, chatGptResp),
+			fmt.Sprintf(`{"clearedAt":"%s","errorCodes":%s}`, null.Time{}.Time.UTC().Format(time.RFC3339Nano), chtJSON),
```
elffjs, line 366: I don't think having the zero time appear here is helpful.

I'm going to change this.

internal/controllers/device_data_controller.go
```diff
@@ -407,6 +408,47 @@ func (udc *UserDevicesController) GetUserDeviceErrorCodeQueries(c *fiber.Ctx) er
...
+	if err != nil {
+		logger.Err(err).Msg("error occurred when fetching error codes for device")
+		return fiber.NewError(fiber.StatusInternalServerError, "error occurred fetching device error queries")
+	}
+
+	if errCodeQuery.ClearedAt.Valid {
+		return fiber.NewError(fiber.StatusBadRequest, "all error codes already cleared")
+	}
+
+	errCodeQuery.ClearedAt = null.TimeFrom(time.Now().UTC().Truncate(time.Microsecond))
```
elffjs, line 436: Postgres already cuts down to a microsecond, no? Is this for tests?

internal/controllers/device_data_controller_test.go
```diff
@@ -726,3 +727,156 @@ func TestUserDevicesController_GetUserDevicesErrorCodeQueries(t *testing.T) {
...
+			err = erCodeQuery.Insert(ctx, pdb.DBS().Writer, boil.Infer())
+			assert.NoError(t, err)
+		}
+
+		request := test.BuildRequest("POST", fmt.Sprintf("/user/devices/%s/error-codes/clear", ud.ID), "")
+		response, _ := app.Test(request)
+		body, _ := io.ReadAll(response.Body)
+
+		assert.Equal(t, fiber.StatusBadRequest, response.StatusCode)
+		assert.Equal(t, string(body), "all error codes already cleared")
```
elffjs, line 877: Note that these are `assert.X(t, expected, actual)`

elffjs:
Sorry, I got distracted by other stuff—the one thing I'd like to change here is having the API return null (or just omit) the `clearedAt` field when the query hasn't been cleared.

internal/controllers/device_data_controller.go
```diff
@@ -398,15 +400,62 @@ func (udc *UserDevicesController) GetUserDeviceErrorCodeQueries(c *fiber.Ctx) er
...
-		queries = append(queries, GetUserDeviceErrorCodeQueriesResponseItem{
+		userDeviceresp := GetUserDeviceErrorCodeQueriesResponseItem{
 			ErrorCodes:  ercJSON,
 			RequestedAt: erc.CreatedAt,
-		})
+		}
+
+		/* if !erc.ClearedAt.Valid {
+			userDeviceresp.ClearedAt =
+		} */
```
elffjs, line 410: What's this?

internal/controllers/device_data_controller.go
```diff
@@ -398,15 +400,62 @@ func (udc *UserDevicesController) GetUserDeviceErrorCodeQueries(c *fiber.Ctx) er
 			return err
 		}
 
-		queries = append(queries, GetUserDeviceErrorCodeQueriesResponseItem{
+		userDeviceresp := GetUserDeviceErrorCodeQueriesResponseItem{
 			ErrorCodes:  ercJSON,
 			RequestedAt: erc.CreatedAt,
```
elffjs, line 405: We talked about populating `ClearedAt` here. Can we do that? `null.Time` has a [`Ptr`](https://pkg.go.dev/github.com/volatiletech/null/v8#Time.Ptr) method that should make this simple.

internal/controllers/device_data_controller.go
elffjs: I don't love it—I think it's a bit noisy—but it works.

internal/controllers/device_data_controller.go
```diff
@@ -398,15 +400,62 @@ func (udc *UserDevicesController) GetUserDeviceErrorCodeQueries(c *fiber.Ctx) er
...
+	logger := helpers.GetLogger(c, udc.log)
+
+	errCodeQuery, err := models.ErrorCodeQueries(
+		models.ErrorCodeQueryWhere.UserDeviceID.EQ(udi),
+		qm.OrderBy(models.ErrorCodeQueryColumns.CreatedAt+" DESC"),
+		qm.Limit(1),
+	).One(c.Context(), udc.DBS().Reader)
+	if err != nil {
+		logger.Err(err).Msg("error occurred when fetching error codes for device")
+		return fiber.NewError(fiber.StatusInternalServerError, "error occurred fetching device error queries")
```
elffjs, line 436: Can we throw a 4xx if there are no queries? I just like to avoid 5xx if it's something expected.

elffjs (changes requested):
I think the workflow we want to make sure works is this:

1. I query with a list of DTCs for vehicle V
2. I get back that list of DTCs with a description attached to each
3. When I list all of my queries for vehicle V, I see this new query at the top
4. I hit the clear command on V
5. When I list all the queries for V again, that most recent query is marked as "cleared", with a timestamp

elffjs (approved):
I think a test for what the list endpoint looks like after hitting `/clear` would have been good.

But this has gone on a while, let's merge it.

PR #110, opened 2023-05-23
vc-cred-content-type

Allyson-English:
# Proposed Changes
set content type of vc cred before returning

### DIFF
--- a/internal/controllers/nft_controller.go
+++ b/internal/controllers/nft_controller.go
@@ -725,6 +725,6 @@ func (nc *NFTController) GetVinCredential(c *fiber.Ctx) error {
 		return fiber.NewError(fiber.StatusNotFound, "Credential associated with NFT not found.")
 	}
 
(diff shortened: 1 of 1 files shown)


internal/controllers/nft_controller.go
```diff
@@ -725,5 +725,6 @@ func (nc *NFTController) GetVinCredential(c *fiber.Ctx) error {
 		return fiber.NewError(fiber.StatusNotFound, "Credential associated with NFT not found.")
 	}
 
+	c.Set("content-type", fiber.MIMEApplicationJSON)
```
elffjs, line 728:
```suggestion
	c.Set(fiber.HeaderContentType, fiber.MIMEApplicationJSON)
```

elffjs (changes requested):
Silly suggestion.

elffjs (approved):

PR #111, opened 2023-05-24
Pla 1490 virtual device minting

0xdev22:
# Proposed Changes


internal/controllers/virtual_devices_controller.go
```diff
@@ -0,0 +1,210 @@
...
+
+// SignVirtualDeviceMintingPayload godoc
+// @Description validate signed signature for vehicle minting
+// @Tags        integrations
+// @Produce     json
+// @Success     200 {array}
+// @Router      /integration/:tokenID/mint-virtual-device [post]
+func (vc *VirtualDeviceController) SignVirtualDeviceMintingPayload(c *fiber.Ctx) error {
+	tokenID := c.Params("tokenID")
+	vehicleNode := c.Query("vehicle_id")
```
JamesReate, line 148: why would `vehicle_id` not be in the path and instead in the querystring - is it optional?

internal/test/helpers.go
```diff
@@ -515,3 +518,37 @@ func BuildDeviceDefinitionGRPC(deviceDefinitionID string, mk string, model strin
 
 	return []*ddgrpc.GetDeviceDefinitionItemResponse{rp}
 }
+
+func BuildGetUserGRPC(id string, email *string, ethereum_address *string, referred_by *users.UserReferrer) *pb.User {
```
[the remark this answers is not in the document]
elffjs: Similarly, `referredBy`.

internal/controllers/virtual_devices_controller.go
```diff
@@ -0,0 +1,210 @@
...
+	uTokenID, err := strconv.ParseUint(tokenID, 10, 64)
+	if err != nil {
+		return fiber.NewError(fiber.StatusBadRequest, "invalid tokenID provided")
+	}
+
+	user, err := vc.usersClient.GetUser(c.Context(), &pb.GetUserRequest{
+		Id: userID,
+	})
+	if err != nil {
+		vc.log.Debug().AnErr("GetUser Grpc", err).Msg("error occurred when fetching user")
```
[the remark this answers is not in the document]
elffjs: Let's use the gRPC error translator. There is a difference between "can't find you" and "I can't reach the upstream service". It's not super precise but it's better than no distinction.

internal/controllers/virtual_devices_controller.go
```diff
@@ -0,0 +1,210 @@
...
+		return fiber.NewError(fiber.StatusUnauthorized, "error occurred when fetching user")
+	}
+
+	if user.EthereumAddress == nil {
+		return fiber.NewError(fiber.StatusUnauthorized, "User does not have an Ethereum address on file.")
+	}
+
+	integration, err := vc.deviceDefSvc.GetIntegrationByTokenID(c.Context(), uTokenID)
+	if err != nil {
+		return fiber.NewError(fiber.StatusBadRequest, "failed to get integration")
```
elffjs, line 114: Translate again. Note that we're just using this to check that the integration _exists_.

internal/controllers/virtual_devices_controller.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return fiber.NewError(fiber.StatusBadRequest, "invalid tokenID provided")
+	}
+
+	integration, err := vc.deviceDefSvc.GetIntegrationByTokenID(c.Context(), uTokenID)
+	if err != nil {
+		return helpers.GrpcErrorToFiber(err, "failed to get integration")
+	}
+
+	user, err := vc.usersClient.GetUser(c.Context(), &pb.GetUserRequest{
```
elffjs, line 175: Do we check that this person owns the NFT? That it exists?

I think we could consolidate the following logic:

1. User on file
2. User has address
3. Vehicle NFT exists
4. User owns NFT

I'm sure it's annoying to write twice.

elffjs: Let's store `userAddr = common.HexToAddress(*user.EthereumAddress)`.

internal/controllers/virtual_devices_controller.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return fiber.NewError(fiber.StatusInternalServerError, "Couldn't not verify signature.")
+	}
+
+	pubKey, err := crypto.UnmarshalPubkey(sigPublicKey)
+	if err != nil {
+		return fiber.NewError(fiber.StatusInternalServerError, "Couldn't not verify signature.")
+	}
+
+	payloadVerified := bytes.Equal(crypto.PubkeyToAddress(*pubKey).Bytes(), common.HexToAddress(*user.EthereumAddress).Bytes())
```
elffjs, line 205: Then you can just check `crypto.PubkeyToAddress(*pubKey) = userAddr`.

internal/controllers/virtual_devices_controller.go
elffjs: 3. `POST virtual-device/mint/{integrationId}/{vehicleId}` (which order?) is also a possibility. The advantage here is that it's easy to search in Nginx, which is powerful.

0xdev22: Good point, will make the change

elffjs: I am iffy on this as a URL pattern.

internal/controllers/virtual_devices_controller.go
```diff
@@ -0,0 +1,233 @@
...
+	"github.com/savsgio/gotils/bytes"
+	"github.com/volatiletech/null/v8"
+	"github.com/volatiletech/sqlboiler/v4/types"
+)
+
+type VirtualDeviceController struct {
+	Settings     *config.Settings
+	DBS          func() *db.ReaderWriter
+	log          *zerolog.Logger
+	integSvc     services.DeviceDefinitionIntegrationService
```
elffjs, line 30: Do we use this for anything?

internal/controllers/virtual_devices_controller.go
```diff
@@ -0,0 +1,233 @@
...
+		return fiber.NewError(fiber.StatusBadRequest, "Couldn't verify signature.")
+	}
+
+	pubRaw, err := crypto.UnmarshalPubkey(pub)
+	if err != nil {
+		vc.log.Err(err).Msg("Error occurred marshalling recovered public public key")
+		return fiber.NewError(fiber.StatusBadRequest, "Couldn't verify signature.")
+	}
+
+	payloadVerified := bytes.Equal(crypto.PubkeyToAddress(*pubRaw).Bytes(), userAddr.Bytes())
```
elffjs, line 226: These are both `common.Address`; you can do `crypto.PubkeyToAddress(*pubRaw) == userAddr`.

internal/controllers/virtual_devices_controller.go
```diff
@@ -0,0 +1,233 @@
...
+	deviceDefSvc services.DeviceDefinitionService
+	usersClient  pb.UserServiceClient
+}
+
+type SignVirtualDeviceMintingPayloadRequest struct {
+	VehicleNode int `json:"vehicleNode"`
+	Credentials struct {
+		AuthorizationCode string `json:"authorizationCode"`
+	} `json:"credentials"`
+	OwnerSignature string `json:"ownerSignature"`
```
elffjs, line 40: Let's use `hexutil.Bytes`. That saves us the work of having to parse this out.

internal/controllers/virtual_devices_controller.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// GetVirtualDeviceMintingPayload godoc
+// @Description gets the payload for to mint virtual device given an integration token ID
+// @Tags        integrations
+// @Produce     json
+// @Param       integrationNode path int true "token ID"
+// @Param       vehicleID path int true "vehicle ID"
+// @Success     200 {array} signer.TypedData
+// @Router      /mint/:integrationNode/:vehicleID [get]
```
elffjs, line 113: I think we want the full path after `v1` here.

internal/controllers/virtual_devices_controller.go
```diff
@@ -0,0 +1,233 @@
...
+	})
+	if err != nil {
+		vc.log.Debug().Err(err).Msg("error occurred when fetching user")
+		return helpers.GrpcErrorToFiber(err, "error occurred when fetching user")
+	}
+
+	vid, err := strconv.ParseInt(vehicleNode, 10, 64)
+	if err != nil {
+		return fiber.NewError(fiber.StatusBadRequest, "invalid vehicleNode provided")
+	}
```
elffjs, line 135: Can we consolidate the parsing?

internal/controllers/virtual_devices_controller.go
```diff
@@ -0,0 +1,233 @@
...
+
+	if err = vc.verifyUserAddressAndNFTExist(c.Context(), user, vid, rawIntegrationNode); err != nil {
+		return err
+	}
+
+	integration, err := vc.deviceDefSvc.GetIntegrationByTokenID(c.Context(), integrationNode)
+	log.Println("avbbbbbbbbb", integration)
```
elffjs, line 142: I am going to remove these.

internal/controllers/virtual_devices_controller.go
elffjs: Eh it's too strict, I forgot that it requires `0x` in the front.

internal/controllers/virtual_devices_controller.go
elffjs: Deleted it.

PR #112, opened 2023-05-24
PLA-1501-autopi-access-middleware

Allyson-English:
Protect the AutoPi endpoints with middleware that enforces the following:

- If the AutoPi is unpaired (on-chain), then anyone can access
- If the AutoPi is paired, then the user must have an address on file, and that address must either be the owner of the AutoPi or the owner of the paired vehicle


cmd/devices-api/api.go
```diff
@@ -194,12 +194,17 @@ func startWebAPI(logger zerolog.Logger, settings *config.Settings, pdb db.Store,
...
-	v1Auth.Get("/autopi/unit/:unitID", userDeviceController.GetAutoPiUnitInfo)
-	v1Auth.Get("/autopi/unit/:unitID/is-online", userDeviceController.GetIsAutoPiOnline)
+
+	// Autopi specific routes.
+	apOwnerMw := owner.AutoPi(pdb, usersClient, &logger)
+	apOwner := v1Auth.Group("/autopi/unit/:unitID", apOwnerMw)
+
+	apOwner.Get("/", userDeviceController.GetAutoPiUnitInfo)
+	apOwner.Get("/is-online", userDeviceController.GetIsAutoPiOnline)
+	apOwner.Post("/update", userDeviceController.StartAutoPiUpdateTask)
```
elffjs, line 204: I think before adding the middleware here we should check whether the bodies of the handlers are undoing our hard work. And also to use our logger, etc.

internal/middleware/owner/owner.go
```diff
@@ -15,14 +20,14 @@ import (
 
 var errNotFound = fiber.NewError(fiber.StatusNotFound, "Device not found.")
 
-// New creates a new middleware handler that checks whether a user is authorized to access
+// UserDeviceOwner creates a new middleware handler that checks whether a user is authorized to access
```
elffjs, line 23: Out of date?

elffjs, line 23:
```suggestion
// UserDevice creates a new middleware handler that checks whether a user is authorized to access
```

internal/middleware/owner/owner.go
```diff
@@ -67,3 +72,58 @@ func New(dbs db.Store, usersClient pb.UserServiceClient, logger *zerolog.Logger)
 		return errNotFound
 	}
 }
+
+// AutoPiOwner creates a new middleware handler that checks whether an autopi is paired.
```
elffjs, line 76:
```suggestion
// AutoPi creates a new middleware handler that checks whether an autopi is paired.
```

internal/middleware/owner/owner.go
```diff
@@ -67,3 +72,58 @@ func New(dbs db.Store, usersClient pb.UserServiceClient, logger *zerolog.Logger)
...
+func AutoPi(dbs db.Store, usersClient pb.UserServiceClient, logger *zerolog.Logger) fiber.Handler {
+	return func(c *fiber.Ctx) error {
+		userID := helpers.GetUserID(c)
+		unitID := c.Params("unitID")
+		valid, unitID := services.ValidateAndCleanUUID(unitID)
+		if !valid {
+			return fiber.NewError(fiber.StatusBadRequest, "Unit id is not a valid UUID.")
+		}
+
+		logger := logger.With().Str("userId", userID).Str("unitID", unitID).Logger()
```
elffjs, line 93:
```suggestion
		logger := logger.With().Str("userId", userID).Str("unitId", unitID).Logger()
```

internal/middleware/owner/owner.go
```diff
@@ -67,3 +72,58 @@ func New(dbs db.Store, usersClient pb.UserServiceClient, logger *zerolog.Logger)
...
+			}
+			logger.Err(err).Msg("Database failure searching for AutoPi.")
+			return fiber.NewError(fiber.StatusInternalServerError, "Internal error.")
+		}
+
+		// if unit has not been paired or has been unpaired
+		if autopiUnit.PairRequestID.IsZero() || !autopiUnit.UnpairRequestID.IsZero() {
+			logger.Info().Msg("AutoPi is not paired, anyone can access endpoint.")
+			return nil
+		}
```
elffjs, line 112: Could we just look for the presence of a vehicle token id? I want to slowly phase these fields out, they're messy.

internal/middleware/owner/owner.go
```diff
@@ -67,3 +72,58 @@ func New(dbs db.Store, usersClient pb.UserServiceClient, logger *zerolog.Logger)
...
+
+		usr, err := usersClient.GetUser(context.Background(), &pb.GetUserRequest{Id: userID})
+		if err != nil {
+			return err
```
elffjs, line 116: It would be nice to separate out "user not found" from "5xx error" but it shouldn't happen, this is fine.

internal/middleware/owner/owner.go
```diff
@@ -67,3 +72,58 @@ func New(dbs db.Store, usersClient pb.UserServiceClient, logger *zerolog.Logger)
...
+		}
+
+		userAddr := common.HexToAddress(usr.GetEthereumAddress())
```
elffjs, line 119: What if `usr.EthereumAddress` is `nil`? In which case, this will return `""`.

internal/middleware/owner/owner.go
```diff
@@ -67,3 +72,58 @@ func New(dbs db.Store, usersClient pb.UserServiceClient, logger *zerolog.Logger)
...
+				return fiber.NewError(fiber.StatusNotFound, "AutoPi not minted, or unit ID invalid.")
+			}
+			logger.Err(err).Msg("Database failure searching for AutoPi.")
+			return fiber.NewError(fiber.StatusInternalServerError, "Internal error.")
+		}
+
+		// if unit has not been paired or has been unpaired
+		if autopiUnit.PairRequestID.IsZero() || !autopiUnit.UnpairRequestID.IsZero() {
+			logger.Info().Msg("AutoPi is not paired, anyone can access endpoint.")
+			return nil
```
elffjs, line 111: I think we want `c.Next()` here. Maybe put a basic handler that actually returns something in the tests, and check the body? Annoying to do.

internal/middleware/owner/owner.go
```diff
@@ -67,3 +72,58 @@ func New(dbs db.Store, usersClient pb.UserServiceClient, logger *zerolog.Logger)
...
+		}
+
+		usr, err := usersClient.GetUser(context.Background(), &pb.GetUserRequest{Id: userID})
+		if err != nil {
+			return err
+		}
+
+		userAddr := common.HexToAddress(usr.GetEthereumAddress())
+
+		if (userAddr != common.BytesToAddress(autopiUnit.EthereumAddress.Bytes)) &&
```
elffjs, line 121: The `autopi_units.ethereum_address` is the address for the wallet owned by the AutoPi. I think you want to check the address of the owner of the vehicle NFT, so you may have to join here.

internal/middleware/owner/owner.go
```diff
@@ -67,3 +72,58 @@ func New(dbs db.Store, usersClient pb.UserServiceClient, logger *zerolog.Logger)
...
+			(userAddr != common.BytesToAddress(autopiUnit.OwnerAddress.Bytes)) {
+			logger.Info().Msg("User is not owner of paired vehicle or AutoPi.")
+			return errors.New("user address does not match")
```
elffjs, line 124: Some `fiber.NewError(fiber.StatusForbidden, …)`, no?

internal/middleware/owner/owner.go
```diff
@@ -67,3 +72,58 @@ func New(dbs db.Store, usersClient pb.UserServiceClient, logger *zerolog.Logger)
...
+		}
+
+		return nil
```
elffjs, line 127: Same concern here.

elffjs (changes requested):

internal/middleware/owner/owner.go
Allyson-English: wouldn't that give us a false positive if someone unpaired the device?

elffjs: Then we clear that field!

elffjs (commented):
Looks better, at a glance. Let me take out some of the outdated logic in the handlers and then we can work on those.

PR #116, opened 2023-06-02
Pro 1520 create submit calldata

0xdev22:
# Proposed Changes


elffjs:
Can you target this at the other PR?

elffjs:
I tried to rename everything I could find, but I probably missed a few. Please keep an eye out.

internal/controllers/synthetic_devices_controller.go
```diff
@@ -2,14 +2,20 @@ package controllers
 
 import (
 	"context"
+	"math/big"
+	"math/rand"
 	"strconv"
+	"time"
 
 	"github.com/DIMO-Network/devices-api/internal/config"
+	registryContract "github.com/DIMO-Network/devices-api/internal/contracts"
```
elffjs, line 11: Can we take off this rename? It's long and looks like a variable name. Just see how it looks.

internal/controllers/synthetic_devices_controller.go
```diff
@@ -219,10 +233,59 @@ func (vc *SyntheticDevicesController) MintSyntheticDevice(c *fiber.Ctx) error {
...
+	err = vc.sendVirtualDeviceMintPayload(c.Context(), hash, req.VehicleNode, integration.TokenId, userAddr, signature)
+	if err != nil {
+		vc.log.Err(err).Msg("virtual device minting request failed")
+		return fiber.NewError(fiber.StatusInternalServerError, "virtual device minting request failed")
+	}
+
+	return c.Send([]byte("virtual device mint request successful"))
+}
+
+func (vc *SyntheticDevicesController) sendVirtualDeviceMintPayload(ctx context.Context, hash []byte, vehicleNode int, intTokenID uint64, userAddr common.Address, signature []byte) error {
```
elffjs, line 249: Can we call `signature` `ownerSignature`? Just to distinguish it.

internal/controllers/synthetic_devices_controller_test.go
```diff
@@ -209,6 +214,16 @@ func (s *VirtualDevicesControllerTestSuite) TestSignVirtualDeviceMintingPayload(
...
 
+	s.virtDeviceSigSvc.EXPECT().SignHash(gomock.Any(), gomock.Any(), gomock.Any())
+
+	var kb []byte
+	mockProducer.ExpectSendMessageWithCheckerFunctionAndSucceed(func(val []byte) error {
+		kb = val
+		return nil
+	})
+
+	log.Println(string(kb))
```
elffjs, line 225: Let's delete this, but check some stuff:

* It's a CloudEvent of the right type
* It's [targeted at the right address](https://github.com/DIMO-Network/devices-api/blob/2aaf10bf377c3b3fbf0edd36032111cf6433b4b9/internal/services/registry/client.go#L34)
* If we [decode the calldata](https://github.com/DIMO-Network/devices-api/blob/2aaf10bf377c3b3fbf0edd36032111cf6433b4b9/internal/controllers/user_integrations_controller_test.go#L881), it's a call to the right function, with the right inputs. You'll probably have to play with this `.Unpack` thing.

I think it's fair to say that the abstraction of `registry.Client` is not great. That's my fault.

elffjs (changes requested):

internal/controllers/synthetic_devices_controller.go
```diff
@@ -219,10 +233,59 @@ func (vc *SyntheticDevicesController) MintSyntheticDevice(c *fiber.Ctx) error {
...
+	if err != nil {
+		vc.log.Err(err).
+			Str("function-name", "SyntheticWallet.SignHash").
+			Bytes("Hash", hash).
+			Int("childKeyNumber", childKeyNumber).
+			Msg("Error occurred signing message hash")
+		return err
+	}
+
+	client := registry.Client{
```
elffjs, line 261: Similarly, we don't need to keep constructing this. Again, my fault elsewhere.

internal/controllers/synthetic_devices_controller_test.go
```diff
@@ -224,7 +257,50 @@ func (s *VirtualDevicesControllerTestSuite) TestSignVirtualDeviceMintingPayload(
...
+	method := abi.Methods["mintVirtualDeviceSign"]
+
+	callData := me.Data.Data
+
+	s.EqualValues(method.ID, callData[:4])
+
+	o, err := method.Inputs.Unpack(callData[4:])
+	s.Require().NoError(err)
+
+	actualMnInput := o[0].(struct {
```
elffjs, line 282: This is interesting. I was wondering how it would handle the struct.

internal/controllers/synthetic_devices_controller_test.go
```diff
@@ -224,7 +257,50 @@ func (s *VirtualDevicesControllerTestSuite) TestSignVirtualDeviceMintingPayload(
...
+		IntegrationNode   *big.Int       "json:\"integrationNode\""
```
elffjs, line 283: Could we write these as, e.g., `json:"integrationNode"`?

.gitignore
```diff
@@ -24,4 +24,6 @@ data/ipfs/
 dist/
 .idea/
 # ignore settings file as it could contain secrets
-settings.yaml
\ No newline at end of file
+settings.yaml
+
+test.json
```
elffjs, line 29: What's this?

internal/controllers/synthetic_devices_controller_test.go
```diff
@@ -135,7 +157,7 @@ func (s *VirtualDevicesControllerTestSuite) TestGetVirtualDeviceMintingPayload_U
 
 func (s *VirtualDevicesControllerTestSuite) TestGetVirtualDeviceMintingPayload_NoEthereumAddressForUser() {
 	email := "some@email.com"
-	user := test.BuildGetUserGRPC(testUserID, &email, nil, &users.UserReferrer{})
+	user := test.BuildGetUserGRPC(mockUserID, &email, nil, &users.UserReferrer{})
```
elffjs, line 160: It's difficult to believe that this is the problem.

Is this the janky users API mock? Maybe we should use a "real" one, I wonder if stuff is sharing between sessions.

internal/controllers/synthetic_devices_controller.go
```diff
@@ -140,13 +152,13 @@ func (vc *SyntheticDevicesController) GetSyntheticDeviceMintingMessage(c *fiber.
...
 
-	response := vc.getVirtualDeviceMintPayload(int64(integration.TokenId), vid)
+	response := vc.getEIP712(int64(integration.TokenId), vid)
 
 	return c.JSON(response)
 }
 
 // MintSyntheticDevice godoc
-// @Description Validate and submit to the blockchain a synthetic device minting request.
+// @Description Submit a metadata
```
elffjs, line 161: Did I garble this?

elffjs (changes requested):

internal/services/synthetic_device_instance_service.go
```diff
@@ -0,0 +1,75 @@
...
+		grpc:        grpc,
+	}, nil
+}
+
+func (v *syntheticWalletInstanceService) GetAddress(ctx context.Context, childNumber uint32) ([]byte, error) {
+	res, err := v.grpc.client.GetAddress(ctx, &pb.GetAddressRequest{
+		ChildNumber: childNumber,
+	})
+
+	defer v.grpc.conn.Close()
```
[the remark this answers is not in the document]
elffjs: My prediction was correct!

.gitignore
0xdev22: It gets generated when you run all tests using the command used in the github actions

PR #117, opened 2023-06-05
PRO-1526-internal-vin-endpoint

Allyson-English:
- Internal gRPC endpoint, accessible by admin
- Allows an admin to issue a VIN VC for a vehicle NFT with a prescribed, finite length


internal/services/issuer/verifible_credential.go
```diff
@@ -74,10 +74,13 @@ func New(c Config) (*Issuer, error) {
 	}, nil
 }
 
-func (i *Issuer) VIN(vin string, tokenID *big.Int) (id string, err error) {
+// VIN issues a vin credential using device vin and token ID
+// if an empty string is passed for expirationDate, the expiration date
+// defaults to eight days from current day.
+func (i *Issuer) VIN(vin string, tokenID *big.Int, expirationDate time.Time) (id string, err error) {
```
elffjs, line 80: It seems bad that we're not using this new argument, yeah?

elffjs (changes requested):
I think we should use the expiry!

pkg/grpc/user_devices.proto
```diff
@@ -26,6 +26,7 @@ service UserDeviceService {
       returns (RegisterUserDeviceFromVINResponse);
   rpc UpdateDeviceIntegrationStatus(UpdateDeviceIntegrationStatusRequest)
       returns (UserDevice);
+  rpc IssueVinCredential(VinCredentialRequest) returns (VinCredentialResponse);
```
elffjs, line 29: I think we should draw some distinction in the request/response here that this is for the issuing endpoint. Like `VinCredentialResponse` could be confused with `VinCredential`. I got them confused!

elffjs:
And it’s out of sync again. I’ll fix this up.

PR #121, opened 2023-06-07
Pro 1544 populate synthetic device table

0xdev22:
# Proposed Changes


internal/controllers/synthetic_devices_controller.go
```diff
@@ -235,25 +238,39 @@ func (vc *SyntheticDevicesController) MintSyntheticDevice(c *fiber.Ctx) error {
 		return fiber.NewError(fiber.StatusBadRequest, "Invalid signature provided")
 	}
 
-	err = vc.sendVirtualDeviceMintPayload(c.Context(), hash, req.VehicleNode, integration.TokenId, ownerSignature)
+	childKeyNumber := generateRandomNumber()
```
elffjs, line 241: I think we want to stop using this. I don't think there's a point to using `IDENTITY` if we're generating random stuff in Go. That does put us in a bit of a bind if we want the address to be non-null.

elffjs: I would use [`CREATE SEQUENCE`](https://www.postgresql.org/docs/current/sql-createsequence.html), how about that?

internal/controllers/synthetic_devices_controller.go
```diff
@@ -235,25 +238,39 @@ func (vc *SyntheticDevicesController) MintSyntheticDevice(c *fiber.Ctx) error {
...
 
-	return c.Send([]byte("virtual device mint request successful"))
-}
+	vnID := types.NewDecimal(decimal.New(vid, 0))
+	syntheticDevice := &models.SyntheticDevice{
+		VehicleTokenID:    vnID,
+		IntegrationID:     integration.Id,
+		WalletChildNumber: childKeyNumber,
+		WalletAddress:     syntheticDeviceAddr,
+		MintRequestID:     vehicleNFT.MintRequestID,
```
elffjs, line 255: We're creating a _new_ meta-transaction for the virtual device, so we need to construct that row first.

elffjs (changes requested):

internal/controllers/synthetic_devices_controller.go
```diff
@@ -235,28 +237,64 @@ func (vc *SyntheticDevicesController) MintSyntheticDevice(c *fiber.Ctx) error {
 		return fiber.NewError(fiber.StatusBadRequest, "Invalid signature provided")
 	}
 
-	// Yikes!
-	ownerSignature[64] += 27
```
elffjs, line 239: Can we get away with deleting this? I don't think it's wise.

internal/controllers/synthetic_devices_controller.go
```diff
@@ -235,28 +237,64 @@ func (vc *SyntheticDevicesController) MintSyntheticDevice(c *fiber.Ctx) error {
...
-func (vc *SyntheticDevicesController) sendVirtualDeviceMintPayload(ctx context.Context, hash []byte, vehicleNode int, intTokenID uint64, ownerSignature []byte) error {
-	childKeyNumber := generateRandomNumber()
+	requestID := ksuid.New().String()
+	metaReq := &models.MetaTransactionRequest{
+		ID:     requestID,
+		Status: models.MetaTransactionRequestStatusUnsubmitted,
+	}
 
+	if err = metaReq.Insert(context.Background(), tx, boil.Infer()); err != nil {
+		tx.Rollback() // nolint
```
elffjs, line 265: Just `defer tx.Rollback()` after the transaction gets created successfully.

internal/controllers/synthetic_devices_controller.go
```diff
@@ -280,12 +318,17 @@ func (vc *SyntheticDevicesController) sendVirtualDeviceMintPayload(ctx context.C
...
-	return vc.registryClient.MintVirtualDeviceSign(requestID, mvt)
+	return syntheticDeviceAddr, vc.registryClient.MintVirtualDeviceSign(requestID, mvt)
 }
 
-func generateRandomNumber() int {
-	rand.Seed(time.Now().UnixNano())
-	min := 1
-	max := 1000
-	return rand.Intn(max-min+1) + min
+func (vc *SyntheticDevicesController) generateRandomNumber(ctx context.Context) (*int, error) {
```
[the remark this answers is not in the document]
elffjs: I'm curious—why `*int` and not `int`?

internal/controllers/synthetic_devices_controller.go
```diff
@@ -235,28 +237,64 @@ func (vc *SyntheticDevicesController) MintSyntheticDevice(c *fiber.Ctx) error {
...
+		vc.log.Err(err).Msg("error occurred saving synthetic device")
+		return fiber.NewError(fiber.StatusInternalServerError, "synthetic device minting request failed")
+	}
+
+	tx.Commit() // nolint
+
+	return c.Send([]byte("synthetic device mint request successful"))
+}
+
+func (vc *SyntheticDevicesController) sendVirtualDeviceMintPayload(ctx context.Context, hash []byte, vehicleNode int, intTokenID uint64, ownerSignature []byte, childKeyNumber int) ([]byte, error) {
```
elffjs, line 290: This signature seems awkward, but I don't have a better suggestion for you.

internal/controllers/synthetic_devices_controller.go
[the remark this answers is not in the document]
elffjs: If your client doesn't check for the error then he is sunk anyway!

internal/controllers/synthetic_devices_controller.go
elffjs: I will add it back.

internal/controllers/synthetic_devices_controller.go
```diff
@@ -235,28 +237,62 @@ func (vc *SyntheticDevicesController) MintSyntheticDevice(c *fiber.Ctx) error {
...
 
-func (vc *SyntheticDevicesController) sendVirtualDeviceMintPayload(ctx context.Context, hash []byte, vehicleNode int, intTokenID uint64, ownerSignature []byte) error {
-	childKeyNumber := generateRandomNumber()
+	requestID := ksuid.New().String()
+	metaReq := &models.MetaTransactionRequest{
+		ID:     requestID,
+		Status: models.MetaTransactionRequestStatusUnsubmitted,
+	}
 
+	if err = metaReq.Insert(context.Background(), tx, boil.Infer()); err != nil {
```
elffjs, line 266: Might as well keep using the request context.

internal/controllers/synthetic_devices_controller.go
```diff
@@ -235,28 +237,62 @@ func (vc *SyntheticDevicesController) MintSyntheticDevice(c *fiber.Ctx) error {
...
+		WalletAddress:     syntheticDeviceAddr,
+		MintRequestID:     requestID,
+	}
+
+	if err = syntheticDevice.Insert(context.Background(), tx, boil.Infer()); err != nil {
+		vc.log.Err(err).Msg("error occurred saving synthetic device")
+		return fiber.NewError(fiber.StatusInternalServerError, "synthetic device minting request failed")
+	}
+
+	tx.Commit() // nolint
```
elffjs, line 285: I think you generally do want to check for the success of `Commit`.

internal/controllers/synthetic_devices_controller.go
```diff
@@ -197,10 +203,6 @@ func (vc *SyntheticDevicesController) MintSyntheticDevice(c *fiber.Ctx) error {
 		return fiber.NewError(fiber.StatusBadRequest, "invalid vehicle id provided")
 	}
 
-	if err = vc.verifyUserAddressAndNFTExist(c.Context(), user, vid, rawIntegrationNode); err != nil {
```
elffjs, line 200: I'm not sure why we got rid of this—this seems like a really important line!

migrations/20230606233841_synthetic_devices.sql
```diff
@@ -0,0 +1,29 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query';
+
+CREATE TABLE synthetic_devices (
+    vehicle_token_id numeric(78, 0) NOT NULL,
+    integration_token_id numeric(78, 0) NOT NULL,
+    mint_request_id char(27) NOT NULL UNIQUE,
+    wallet_child_number int NOT NULL UNIQUE,
```
elffjs, line 9: I'd like to name all these but it's okay, let's just try it.

PR #122, opened 2023-06-08
PRO-1548-reissue-vc-weekly

Allyson-English:
- The first time that we get a status payload on topic.device.status from a vehicle in a given rewards week, carrying the VIN of the vehicle, re-issue the VIN VC
- Do not issue if the VIN in the payload doesn't match the one on file


elffjs (commented):
Looks good at a glance. We should start testing some stuff, probably get rid of this "fraud monitor" that no one looks at.

elffjs:
Let's fix up the tests and the lints, and then talk about tactics.

PR #133, opened 2023-06-15
New GRPC endpoint to update integration status

rhvivancoeffio:
# Proposed Changes


internal/api/user_devices.go
```diff
@@ -267,6 +267,37 @@ func (s *userDeviceService) RegisterUserDeviceFromVIN(ctx context.Context, req *
...
+
+		s.logger.Err(err).Str("IntegrationId", req.IntegrationId).Msg("Database failure retrieving UserDeviceAPIIntegrations.")
+		return nil, status.Error(codes.Internal, "Internal error.")
+	}
+
+	if len(dbDevice.R.UserDeviceAPIIntegrations) == 0 {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("can't find API integration for device %s and integration %s", req.UserDeviceId, req.IntegrationId))
+	}
+
+	apiIntegration := dbDevice.R.UserDeviceAPIIntegrations[0]
```
JamesReate, line 292: there could be more than one integration, ie. for smartcar or autopi. 
Add the `integrationId` to the `req` so you can select the right `dbDevice.R.UserDeviceAPIIntegrations`

internal/api/user_devices.go
```diff
@@ -267,6 +267,37 @@ func (s *userDeviceService) RegisterUserDeviceFromVIN(ctx context.Context, req *
 	return &pb.RegisterUserDeviceFromVINResponse{Created: true}, err
 }
 
+func (s *userDeviceService) UpdateDeviceIntegrationStatus(ctx context.Context, req *pb.UpdateDeviceIntegrationStatusRequest) (*pb.UserDevice, error) {
+	dbDevice, err := models.UserDevices(
+		models.UserDeviceWhere.ID.EQ(req.UserDeviceId),
```
JamesReate, line 272: the `user_device_api_integration` also has the column `user_device_id` so you may be able to simplify and only select from `user_device_api_integration`

JamesReate (commented):
just one simplification when querying the user_device_api_integration

PR #136, opened 2023-06-18
PRO-1598 Use openai functions for openai calls

0xdev22:
# Proposed Changes

### DIFF
--- a/internal/services/openai.go
+++ b/internal/services/openai.go
@@ -29,10 +29,16 @@ type openAI struct {
 	logger     *zerolog.Logger
 }
 
+type FunctionCallResponse struct {
+	Name      string
+	Arguments string
+}
+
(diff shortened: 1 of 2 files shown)


internal/services/openai.go
```diff
@@ -142,21 +182,17 @@ func (o *openAI) GetErrorCodesDescription(make, model string, errorCodes []strin
 		o.logger.Error().Interface("rawResponse", r).Msg("Unexpected finish_reason from ChatGPT.")
 	}
 
-	rawResp := map[string]string{}
-	if err := json.Unmarshal([]byte(r.Choices[0].Message.Content), &rawResp); err != nil {
+	var rawResp ErrorCodesFunctionCallResoponse
```
elffjs, line 185: We could do some sanity checks here but this makes sense.

PR #153, opened 2023-07-12
GRPC - GetAllUserDevice

rhvivancoeffio:
# Proposed Changes


pkg/grpc/user_devices.proto
```diff
@@ -117,4 +118,9 @@ message UpdateDeviceIntegrationStatusRequest {
...
   string integration_id = 2;
   string status = 3;
-}
\ No newline at end of file
+}
+
+message GetAllUserDeviceRequest {
+  string wmi = 1;
+}
+message GetAllUserDeviceResponse { repeated UserDevice user_devices = 1; }
```
JamesReate, line 126: would we want to make this a stream since it could be a lot of data?

internal/api/user_devices.go
```diff
@@ -393,6 +392,35 @@ func (s *userDeviceService) GetAllUserDeviceValuation(ctx context.Context, _ *em
 	}, nil
 }
 
+func (s *userDeviceService) GetAllUserDevice(ctx context.Context, req *pb.GetAllUserDeviceRequest) (*pb.GetAllUserDeviceResponse, error) {
```
JamesReate, line 395: let's got the grpc stream route to be more future proof with db paging - this will be a big call otherwise

JamesReate (approved):

PR #163, opened 2023-07-19
PRO-1653 add manuf token id col, change pkey, update fkeys

JamesReate:
# Proposed Changes
- add device_manufacturer_token_id
- delete bad data
- update pkey
- update and rebuild fkeys


JamesReate:
@elffjs the following test is failing but I can't tell why, the TokenID is nil, which results in a conflict being returned - is this test setup?

elffjs:
I don't know. I will look tomorrow.

JamesReate:
ah it seems to be b/c of the pkey change the boiler Update isn't updating

PR #169, opened 2023-07-26
PRO-1707 - Add powertraintype to grpc endpoint

rhvivancoeffio:
# Proposed Changes


pkg/grpc/aftermarket_devices_grpc.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
-// - protoc-gen-go-grpc v1.3.0
-// - protoc             v4.23.4
+// - protoc-gen-go-grpc v1.2.0
+// - protoc             v3.21.7
```
JamesReate, line 4: update your version of protoc locally

JamesReate (changes requested):
While we're in here it occurred to me let's also add the Protocol:
`CANProtocol *string `json:"canProtocol,omitempty"`` from the UserDeviceMetadata

Then also just update your local protoc to latest and regen protos

PR #174, opened 2023-08-01
MOB-793 autopi handle duplicate VIN for same user

JamesReate:
# Proposed Changes


JamesReate:
there was a failing test but with something unrelated to this code change - although tests did pass for me locally

PR #176, opened 2023-08-01
PRO-1719-synthetic-fingerprint-listener

Allyson-English:
# Proposed Changes

- Listen for synthetic fingerprint events
- If the vin is valid, issue a credential (same process as for device fingerprints)
- Confirm that the vin is valid by checking UserDevices

### Notes

- Currently passing message.Key (which is the user device id) to the handler and using that to query UserDevices table since this seemed like the most straightforward way to validate the vin. This makes the signature different from the handler that we use to process aftermarket device fingerprints, and it would be nice if they were the same.


elffjs:
Oh that's my fault I changed it because it logged all the time on ERC-20 transfers and it drove me insane. Just ignore it, I will fix it.

I appreciate the background here.

internal/services/contracts_events_consumer.go
```diff
@@ -173,9 +173,8 @@ func (c *ContractsEventsConsumer) routeTransferEvent(e *ContractEventData) error
 		return c.handleVehicleTransfer(e)
 	default:
 		c.log.Debug().Str("event", e.EventName).Interface("fullEventData", e).Msg("Handler not provided for contract")
+		return errors.New("Handler not provided for contract")
```
elffjs, line 176: Could you revert this? I'll fix it up on main.

internal/config/settings.go
```diff
@@ -83,6 +83,9 @@ type Settings struct {
 
 	DeviceFingerprintTopic         string `yaml:"DEVICE_FINGERPRINT_TOPIC"`
 	DeviceFingerprintConsumerGroup string `yaml:"DEVICE_FINGERPRINT_CONSUMER_GROUP"`
+
+	SyntheticFingerprintTopic         string `yaml:"SYNTHETIC_FINGERPRINT_TOPIC"`
```
elffjs, line 87: Can we give these some values in `values.yaml`?

internal/services/fingerprint/consumer.go
```diff
@@ -122,7 +147,54 @@ func (c *Consumer) Handle(ctx context.Context, event *Event) error {
...
+		return err
+	}
+
+	c.logger.Info().Str("device-addr", event.Subject).Msg("issued vin credential")
+
+	return nil
+}
+
+func (c *Consumer) HandleSyntheticFingerprint(ctx context.Context, event *Event, key string) error {
+	if !common.IsHexAddress(event.Subject) {
```
elffjs, line 171: Let's look at the topic together—I just re-used the subject from `topic.device.status`, which is just the id from `user_devices`.

elffjs: I'm trying to not make this dependent on synthetic devices being completely live. Maybe we should move to that, sign the message, etc.; but it's going to be difficult.

internal/config/settings.go
```diff
@@ -83,6 +83,9 @@ type Settings struct {
...
+	SyntheticFingerprintConsumerGroup string `yaml:"SYNTHETIC_FINGERPRINT_CONSUMER_GROUP"`
```
elffjs, line 88: Are we using this value?

internal/services/fingerprint/consumer.go
```diff
@@ -122,7 +147,54 @@ func (c *Consumer) Handle(ctx context.Context, event *Event) error {
...
+	observedVIN, err := services.ExtractVIN(event.Data)
+	if err != nil {
+		if err == services.ErrNoVIN {
+			return nil
+		}
+		return fmt.Errorf("couldn't extract VIN: %w", err)
+	}
+
+	ud, err := models.UserDevices(
+		models.UserDeviceWhere.ID.EQ(key),
```
elffjs, line 184: Here I would use the subject. I understand that these are totally divergent.

internal/services/fingerprint/consumer.go
```diff
@@ -137,7 +209,7 @@ func (c *Consumer) Handle(ctx context.Context, event *Event) error {
 		return err
 	}
 
-	c.logger.Info().Msgf("Issued VIN credential for vehicle %d using device %s.", vn.TokenID, addr)
+	c.logger.Info().Str("device-addr", event.Subject).Msg("issued vin credential")
```
elffjs, line 212: The reason, by the way, that I'm using logger fewer fields is that Elastic starts dropping messages because the field count in the day index gets too high. I think it's possible to get around this with configuration but I am way too lazy.

internal/services/fingerprint/consumer.go
```diff
@@ -74,8 +88,19 @@ func RunConsumer(ctx context.Context, settings *config.Settings, logger *zerolog
...
+			if err := deviceGroup.Consume(ctx, []string{settings.DeviceFingerprintTopic}, c); err != nil {
+				logger.Warn().Err(err).Msgf("Consumer group session ended: %s", settings.DeviceFingerprintTopic)
+			}
+			if ctx.Err() != nil {
+				return
+			}
+		}
+	}()
+
+	go func() {
```
elffjs, line 100: Could we try using this? Just spawn two?

https://github.com/DIMO-Network/shared/blob/01c0fa977163f1169266e4521656cb8ec5f2e81c/kafka/group.go#L42

Getting rid of the `switch` above seems good to me.

internal/services/fingerprint/consumer.go
```diff
@@ -122,7 +147,54 @@ func (c *Consumer) Handle(ctx context.Context, event *Event) error {
...
+
+	return nil
+}
+
+func (c *Consumer) HandleSyntheticFingerprint(ctx context.Context, event *Event, key string) error {
+	if !common.IsHexAddress(event.Subject) {
+		return fmt.Errorf("subject %q not a valid address", event.Subject)
+	}
+
+	observedVIN, err := services.ExtractVIN(event.Data)
```
elffjs, line 175: I included `week` in the payload because I was worried about messages coming in around the cutover, combined with clock skew. Does that make any sense?

elffjs (changes requested):

PR #177, opened 2023-08-02
PRO-1720-vcs-in-vehicle-listing

Allyson-English:
# Proposed Changes
- add credential to `user/devices/me` and `user/devices/shared` endpoints

### Notes
- if there's no vin in the credential this won't error, it just won't return a credential-- is that preferable or is it better to return `"vinCredential":{"valid":false}`


internal/controllers/user_devices_controller.go
```diff
@@ -298,6 +299,28 @@ func (udc *UserDevicesController) dbDevicesToDisplay(ctx context.Context, device
...
+					return nil, err
+				}
+				subj, ok := c["credentialSubject"]
+				if ok {
+					v, ok := subj.(map[string]interface{})["vehicleIdentificationNumber"]
+					if ok {
+						credential = &CredentialData{
+							Vin:            v.(string),
+							ExpirationDate: cred.ExpirationDate,
+							IssuanceDate:   cred.IssuanceDate,
```
elffjs, line 316: Can we set `Valid` to `time.Now().Before(cred.ExpirationDate)`? One less if statement.

elffjs (changes requested):

internal/controllers/user_devices_controller.go
```diff
@@ -2234,7 +2224,7 @@ type UserDeviceFull struct {
 	NFT              *NFTData                      `json:"nft,omitempty"`
 	OptedInAt        *time.Time                    `json:"optedInAt"`
 	PrivilegeUsers   []PrivilegeUser               `json:"privilegedUsers"`
-	Credential       *CredentialData               `json:"vinCredential,omitempty"`
+	Credential       *CredentialData               `json:"VINCredential,omitempty"`
```
elffjs, line 2224: Ah no not what I meant. I'll just

PR #178, opened 2023-08-03
PRO-1622 remove device status consumer

JamesReate:
# Proposed Changes


JamesReate:
failing test is currently expected from past state

PR #183, opened 2023-08-15
PRO-1800 - set powertrain from device definition

rhvivancoeffio:
# Proposed Changes

### DIFF
--- a/internal/controllers/user_devices_controller_test.go
+++ b/internal/controllers/user_devices_controller_test.go
@@ -608,8 +608,7 @@ func (s *UserDevicesControllerTestSuite) TestPatchVIN() {
 	//}, nil)
 
 	s.deviceDefSvc.EXPECT().GetDeviceDefinitionByID(gomock.Any(), dd[0].DeviceDefinitionId).Times(1).Return(dd[0], nil)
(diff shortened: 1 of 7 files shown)


JamesReate (approved):
looks good but don't merge yet

internal/controllers/user_devices_controller_test.go
```diff
@@ -597,15 +597,19 @@ func (s *UserDevicesControllerTestSuite) TestPatchVIN() {
...
+	//		{
+	//			VariableID: 126,
+	//			ValueID:    &evID,
+	//		},
+	//	},
+	//}, nil)
+
+	s.deviceDefSvc.EXPECT().GetDeviceDefinitionByID(gomock.Any(), dd[0].DeviceDefinitionId).Times(1).Return(dd[0], nil)
+	powerTrainType := "ICE"
+	s.deviceDefSvc.EXPECT().ConvertPowerTrainStringToPowertrain(gomock.Any()).Return(powerTrainType)
```
JamesReate: instead of gomock.Any() pass in the actual value that the device definition returns for the PowertrainType attribute. You may need to set the device attribute in the `dd[0]` object.

PR #189, opened 2023-08-25
PRO-1881 - User - devices - on create

rhvivancoeffio:
# Proposed Changes


JamesReate:
after discussion, will also add udMd to set powertrain on smartcar create

PR #208, opened 2023-09-18
GetUserDeviceByEthAddr and protos

lauradeng23:
# Proposed Changes


internal/rpc/user_devices.go
```diff
@@ -122,6 +123,50 @@ func (s *userDeviceRPCServer) GetUserDeviceByVIN(ctx context.Context, req *pb.Ge
 	return s.deviceModelToAPI(dbDevice), nil
 }
 
+func (s *userDeviceRPCServer) GetUserDeviceByEthAddr(ctx context.Context, req *pb.GetUserDeviceByEthAddrRequest) (*pb.UserDevice, error) {
+	aftermarketDevice, err := models.FindAftermarketDevice(ctx, s.dbs().Reader, req.EthAddr)
+	if err != nil {
+		log.Printf("Error finding AftermarketDevice: %v", err)
```
JamesReate, line 129: let's use the logger eg. `s.logger.Err(err).Msg(etc)`

internal/rpc/user_devices.go
```diff
@@ -122,6 +123,50 @@ func (s *userDeviceRPCServer) GetUserDeviceByVIN(ctx context.Context, req *pb.Ge
...
+		return nil, err
+	}
+	if aftermarketDevice == nil {
+		return nil, status.Error(codes.NotFound, "No AftermarketDevice found for the given Ethereum address")
+	}
+
+	if aftermarketDevice.R == nil || aftermarketDevice.R.VehicleToken == nil {
```
JamesReate, line 136: this will always be nil because you're not eager loading the VehicleToken relationship on line 127

JamesReate (approved):

PR #213, opened 2023-10-10
SI-2109-grpc-aftermarket-sub-obj

Allyson-English:
[SI-2109](https://linear.app/dimo/issue/SI-2109/redo-vehicle-grpc-aftermarket-sub-object)

# Proposed Changes
- deprecate `aftermarket_device_token_id` and `aftermarket_device_beneficiary_address` in `UserDevice` (`user_devices.proto`) 
- Add an `aftermarket_device` sub-object to `UserDevice`

### Notes
- Since `AftermarketDevice` already existed in `aftermarket_devices.proto` I added some additional optional fields based on the ticket and used this object instead of creating a new one, to avoid a conflict (`serial` was a required field in the existing `AftermarketDevice` object so this change assumes that the macaroon will also have a unit id that can be populated here) 
- for backward compatibility, I used `[deprecated = true]` for the fields we were removing instead of deleting them and reserving the numbers


internal/rpc/user_devices.go
```diff
@@ -415,12 +415,19 @@ func (s *userDeviceRPCServer) deviceModelToAPI(ud *models.UserDevice) *pb.UserDe
 		}
 
 		if amnft := vnft.R.VehicleTokenAftermarketDevice; amnft != nil {
-			out.AftermarketDeviceTokenId = s.toUint64(amnft.TokenID)
-
-			if amnft.Beneficiary.Valid {
-				out.AftermarketDeviceBeneficiaryAddress = amnft.Beneficiary.Bytes
-			} else if amnft.OwnerAddress.Valid {
-				out.AftermarketDeviceBeneficiaryAddress = amnft.OwnerAddress.Bytes
+			out.AftermarketDevice.Serial = amnft.Serial
```
elffjs, line 418: Probably a nil pointer here.

internal/rpc/user_devices.go
```diff
@@ -415,12 +415,19 @@ func (s *userDeviceRPCServer) deviceModelToAPI(ud *models.UserDevice) *pb.UserDe
...
+			out.AftermarketDevice.UserId = &amnft.UserID.String
```
elffjs, line 419: Probably `amnft.UserID.Ptr()`?

pkg/grpc/aftermarket_devices.proto
```diff
@@ -20,4 +20,7 @@ message AftermarketDevice {
 	string serial = 1;
 	optional string user_id = 2;
 	optional bytes owner_address = 3;
+	optional uint64 token_id = 4;
+	optional uint64 manufacturer_token_id = 5;
+	repeated bytes beneficiary = 6;
```
elffjs, line 25:
```suggestion
	bytes beneficiary = 6;
```

pkg/grpc/user_devices.proto
```diff
@@ -56,12 +57,12 @@ message UserDevice {
   optional uint64 token_id = 3;
   optional google.protobuf.Timestamp opted_in_at = 4;
   optional bytes owner_address = 5;
-  optional uint64 aftermarket_device_token_id = 6;
+  optional uint64 aftermarket_device_token_id = 6 [deprecated = true];
```
elffjs, line 60: Good to deprecate. I think we should still populate these in case anyone is using them. rewards-api certainly is.

elffjs: Isn't it tempting to scrap all this and add `optional uint64 aftermarket_device_manufacturer_token_id`?

internal/rpc/user_devices.go
```diff
@@ -415,12 +415,19 @@ func (s *userDeviceRPCServer) deviceModelToAPI(ud *models.UserDevice) *pb.UserDe
 		}
 
 		if amnft := vnft.R.VehicleTokenAftermarketDevice; amnft != nil {
-			out.AftermarketDeviceTokenId = s.toUint64(amnft.TokenID)
```
elffjs, line 418: We can just leave these in.

internal/services/fingerprint/consumer_test.go
```diff
@@ -176,19 +176,19 @@ func (s *ConsumerTestSuite) TestVinCredentialerHandler_DeviceFingerprint() {
...
-			MetaTxTable:     metaTx,
-			VCTable: models.VerifiableCredential{
-				ClaimID:        claimID,
-				Credential:     []byte{},
-				ExpirationDate: time.Now().AddDate(0, 0, -10),
-			},
-			VehicleNFT:        nft,
-			AftermarketDevice: aftermarketDevice,
-		},
+		// {
```
elffjs, line 179: Are these failing?

internal/rpc/user_devices.go
```diff
@@ -415,13 +415,26 @@ func (s *userDeviceRPCServer) deviceModelToAPI(ud *models.UserDevice) *pb.UserDe
...
+				Serial:              amnft.Serial,
+				UserId:              amnft.UserID.Ptr(),
+				TokenId:             s.toUint64(amnft.TokenID),
+				ManufacturerTokenId: s.toUint64(amnft.DeviceManufacturerTokenID),
+			}
+
+			if amnft.OwnerAddress.Valid {
+				out.AftermarketDevice.OwnerAddress = amnft.OwnerAddress.Bytes
 
-			if amnft.Beneficiary.Valid {
```
elffjs, line 420: This is not, strictly speaking, the same logic. But I think it's okay.

internal/rpc/user_devices.go
```diff
@@ -415,13 +415,26 @@ func (s *userDeviceRPCServer) deviceModelToAPI(ud *models.UserDevice) *pb.UserDe
...
-			} else if amnft.OwnerAddress.Valid {
-				out.AftermarketDeviceBeneficiaryAddress = amnft.OwnerAddress.Bytes
+				if amnft.Beneficiary.Valid {
+					out.AftermarketDevice.Beneficiary = amnft.Beneficiary.Bytes
+				} else {
+					out.AftermarketDevice.Beneficiary = amnft.OwnerAddress.Bytes
+				}
 			}
+
+			// these fields have been depricated but are populated for backwards compatability
```
elffjs, line 435:
```suggestion
			// These fields have been deprecated but are populated for backwards compatibility.
```

PR #214, opened 2023-10-10
Add syntheticDevice to UserDevice

0xdev22:
db synth_devices
grpc user_devices list
single user_device

# Proposed Changes


internal/services/fingerprint/consumer_test.go
```diff
@@ -176,7 +176,7 @@ func (s *ConsumerTestSuite) TestVinCredentialerHandler_DeviceFingerprint() {
 			VehicleNFT:        nft,
 			AftermarketDevice: aftermarketDevice,
 		},
-		{
+		/* {
```
elffjs, line 179: Merge in main when you get the chance. We're just going to skip this until we can remove the dependency on the net.

elffjs:
The slot will conflict with Allyson. We will cross that bridge when the other one gets merged.

internal/services/fingerprint/consumer_test.go
```diff
@@ -462,4 +462,4 @@ func (s *ConsumerTestSuite) TestInvalidSignature() {
 	s.Error(err)
 	s.Equal(err.Error(), "invalid signature recovery id")
 	s.NotEqual(recAddr, addr)
-}
+}
```
elffjs, line 465: I'm surprised the linter allows this.

pkg/grpc/user_devices.proto
```diff
@@ -71,6 +71,12 @@ message UserDevice {
   string postal_code = 17;
   string geo_decoded_country = 18;
   string geo_decoded_state_prov = 19;
+  optional SyntheticDevice syntheticDevice = 20;
+}
+
+message SyntheticDevice {
+  int64 tokenId = 1;
```
elffjs, line 78: The protobuf style for fields is snake case, let's do that. `token_id`, `integration_token_id`. I was thinking of the JSON.

elffjs: Is any code populating these fields?

elffjs (changes requested):

internal/rpc/user_devices.go
```diff
@@ -430,6 +430,16 @@ func (s *userDeviceRPCServer) deviceModelToAPI(ud *models.UserDevice) *pb.UserDe
 				Expiration: timestamppb.New(vc.ExpirationDate),
 			}
 		}
+
+		if sd := vnft.R.VehicleTokenSyntheticDevice; sd != nil {
```
elffjs, line 434: Are we loading this relation? Have we tested this?

elffjs: Like, I think you're on the hook for checking all the callers of `deviceModelToAPI` and seeing if they load the relation you want. Is this good? Probably not.

internal/rpc/user_devices.go
```diff
@@ -448,6 +448,16 @@ func (s *userDeviceRPCServer) deviceModelToAPI(ud *models.UserDevice) *pb.UserDe
...
+			stk, _ := sd.VehicleTokenID.Int64()
```
elffjs, line 453: This doesn't look right. This is the token id for the vehicle.

pkg/grpc/user_devices.proto
```diff
@@ -75,6 +75,12 @@ message UserDevice {
   string geo_decoded_country = 18;
   string geo_decoded_state_prov = 19;
   optional AftermarketDevice aftermarket_device = 20;
+  optional SyntheticDevice syntheticDevice = 21;
+}
+
+message SyntheticDevice {
+  int64 token_id = 1;
+  int64 integration_token_id = 2;
```
elffjs, line 83: Should these be `uint64`s?

internal/rpc/user_devices.go
```diff
@@ -448,6 +459,15 @@ func (s *userDeviceRPCServer) deviceModelToAPI(ud *models.UserDevice) *pb.UserDe
 				Expiration: timestamppb.New(vc.ExpirationDate),
 			}
 		}
+
+		if sd := vnft.R.VehicleTokenSyntheticDevice; sd != nil {
```
elffjs, line 463: I think we should be careful here. There is a state in between where the synthetic device is getting minted but has not acquired a token id yet. Can we only populate this when the token id is populated? I think it’s going to be confusing to clients.

Is it going to matter a lot? Probably not, but it will happen.

settings.sample.yaml
```diff
@@ -37,8 +37,6 @@ AUTO_PI_API_URL:
 DOCUMENTS_AWS_ACCESS_KEY_ID: test
 DOCUMENTS_AWS_SECRET_ACCESS_KEY: test
 DOCUMENTS_AWS_ENDPOINT: http://localhost:4566
-NFT_AWS_ACCESS_KEY_ID: test
```
elffjs, line 40: Not used anymore?

0xdev22: Already exists https://github.com/DIMO-Network/devices-api/pull/214/files/e4ffeaa230cc75decdb078e9b88fbf3a90605569#diff-d82ffc9c00e6095149349dcf6b922139639ba720e3b77afcab0e843f4dac4539L25

elffjs: Works for me.

internal/rpc/user_devices.go
```diff
@@ -448,6 +459,17 @@ func (s *userDeviceRPCServer) deviceModelToAPI(ud *models.UserDevice) *pb.UserDe
...
```
elffjs, line 463: Could we do `sd != nil && !sd.TokenID.IsZero()`? Then we could get rid of the nesting below.

internal/rpc/user_devices.go
```diff
@@ -448,6 +460,16 @@ func (s *userDeviceRPCServer) deviceModelToAPI(ud *models.UserDevice) *pb.UserDe
 				Expiration: timestamppb.New(vc.ExpirationDate),
 			}
 		}
+
+		if sd := vnft.R.VehicleTokenSyntheticDevice; sd != nil && !sd.TokenID.IsZero() {
+			log.Println(sd.TokenID.IsZero(), sd != nil, sd != nil && !sd.TokenID.IsZero())
```
elffjs, line 465: Get rid of this?

elffjs (approved):

elffjs:
@Allyson-English Could you take a quick look?

Allyson-English (approved):

PR #219, opened 2023-10-13
Remove loggers when re-assign AutoPi template

Khalaro:
GetVehicleLoggers(vehicleID) which calls GET /obd/loggers/?vehicle=XX

DeleteVehicleLogger which calls `DELETE /obd/loggers/<logger_type>/xxxID/

ONLY when logger's "parent" field is empty, which indicates the logger is not associated by template

# Proposed Changes


internal/services/autopi_api_service.go
```diff
@@ -160,6 +162,30 @@ func (a *autoPiAPIService) UnassociateDeviceTemplate(deviceID string, templateID
 	return nil
 }
 
+// GetVehicleLoggers Get slice of loggers from vehicleID, populates parameter "loggers *[]AutoPiVehicleLogger"
+func (a *autoPiAPIService) GetVehicleLoggers(vehicle AutoPiDongleVehicle, loggers *[]AutoPiVehicleLogger) error {
```
JamesReate, line 166: Change the input parameter to be just the autopiVehicleID. No need to pass in the entire object since in the code all you're using is the `vehicle.ID` field. 
Also, change the ` loggers *[]AutoPiVehicleLogger` from being an input parameter to being returned, eg. `(*[]AutoPiVehicleLogger, error), and then on line 176 just return the loggers.

internal/services/autopi/hardware_template_service.go
```diff
@@ -151,6 +152,24 @@ func (a *hardwareTemplateService) ApplyHardwareTemplate(ctx context.Context, req
...
+		vehicleLoggers := new([]services.AutoPiVehicleLogger)
+		err = a.ap.GetVehicleLoggers(autoPi.Vehicle, vehicleLoggers)
+		if err != nil {
+			return nil, errors.Wrap(err, fmt.Sprintf("failed to get vehicle loggers for vehicleID %d", autoPi.Vehicle.ID))
+		} else {
+			for _, logger := range *vehicleLoggers {
+				if logger.Parent == 0 {
+					loggerType := logger.LoggerType
+					loggerID := logger.ID
+					err := a.ap.DeleteVehicleLogger(loggerType, loggerID)
```
JamesReate, line 165: just inline loggerType and loggerID from above two lines, eg. `(logger.LoggerType, logger.ID)`

internal/services/autopi/hardware_template_service.go
```diff
@@ -151,6 +152,24 @@ func (a *hardwareTemplateService) ApplyHardwareTemplate(ctx context.Context, req
...
 			return nil, errors.Wrap(err, "failed to unassociate template")
 		}
+
+		vehicleLoggers := new([]services.AutoPiVehicleLogger)
+		err = a.ap.GetVehicleLoggers(autoPi.Vehicle, vehicleLoggers)
+		if err != nil {
+			return nil, errors.Wrap(err, fmt.Sprintf("failed to get vehicle loggers for vehicleID %d", autoPi.Vehicle.ID))
+		} else {
+			for _, logger := range *vehicleLoggers {
+				if logger.Parent == 0 {
```
JamesReate, line 162: what does logger.Parent == 0 mean? When looking at loggers returned for my car that were supposed to be removed, they had a parent > 0.

JamesReate: saw your comment in the PR description, that parent = 0 means not associated by template - is this something AutoPi said?

JamesReate: leave a comment on line 162 explaining why the check for parent == 0. This is usually a good use case for comments in code when there is some otherwise unknown magic required logic.

Khalaro: when not associated by template, parent is actually NULL, but since integers cannot be NULL in golang, when converting from the JSON response, the expectation is that "parent" field would populate as value 0. 
I'll update the commit, no problem

Khalaro:
@JamesReate added changes and comments as requested. 
Ready for final approval

internal/services/autopi/hardware_template_service.go
```diff
@@ -149,8 +150,26 @@ func (a *hardwareTemplateService) ApplyHardwareTemplate(ctx context.Context, req
...
+
+		var vehicleLoggers *[]services.AutoPiVehicleLogger
+		vehicleLoggers, err = a.ap.GetVehicleLoggers(autoPi.Vehicle.ID)
+		if err != nil {
+			return nil, errors.Wrap(err, fmt.Sprintf("failed to get vehicle loggers for vehicleID %d", autoPi.Vehicle.ID))
+		} else {
+			for _, logger := range *vehicleLoggers {
+				if logger.Parent == 0 { // IF (logger was not assigned via template)
+					// Autopi's API returns a NULL value for "AutoPiVehicleLogger.Parent" when the logger was not assigned via template
+					// Since integers in Golang cannot be NIL, this will be represented as value 0 after converting from JSON
```
JamesReate, line 164: btw integers in go can be nil if you make them a pointer, eg. `Parent *int`

internal/services/autopi/hardware_template_service.go
```diff
@@ -149,8 +150,26 @@ func (a *hardwareTemplateService) ApplyHardwareTemplate(ctx context.Context, req
 	if autoPi.Template > 0 {
 		err = a.ap.UnassociateDeviceTemplate(autoPi.ID, autoPi.Template)
 		if err != nil {
-			return nil, errors.Wrap(err, "failed to unassociate template")
+			return nil, errors.Wrap(err, fmt.Sprintf("failed to unassociate template %d on device %s", autoPi.Template, autoPi.ID))
 		}
+
+		var vehicleLoggers *[]services.AutoPiVehicleLogger
```
JamesReate, line 156: remove unnecessary line to instantiate vehicleLoggers, just assign it inline on next line.

internal/services/autopi/hardware_template_service.go
Khalaro: Changed integers to integer pointers

internal/services/autopi/hardware_template_service.go
Khalaro: done

PR #221, opened 2023-10-25
Block Tesla integration if there are any existing integrations

elffjs:
# Proposed Changes


JamesReate (approved):
lgtm if we wanna move forward with this

PR #224, opened 2023-11-07
SM-270 - fingerprint ingest

rhvivancoeffio:
# Proposed Changes

### DIFF
--- a/internal/appmetrics/counters.go
+++ b/internal/appmetrics/counters.go
@@ -74,8 +74,8 @@ var (
 	FingerprintRequestCount = promauto.NewCounterVec(
 		prometheus.CounterOpts{
 			Name: "devices_api_fingerprint_request_count",
-			Help: "The total number of requests by the Fingerprint",
+			Help: "The total number of Fingerprint requests",
 		},
-		[]string{"protocol"},
+		[]string{"protocol", "status"},
 	)
 )
--- a/internal/appmetrics/counters.go
+++ b/internal/appmetrics/counters.go
@@ -70,4 +70,12 @@ var (
 		},
 		[]string{"method", "path", "status"},
 	)
+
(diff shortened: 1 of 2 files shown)


internal/appmetrics/counters.go
```diff
@@ -70,4 +70,12 @@ var (
 		},
 		[]string{"method", "path", "status"},
 	)
+
+	FingerprintRequestCount = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "devices_api_fingerprint_request_count",
+			Help: "The total number of requests by the Fingerprint",
```
JamesReate, line 77:
```suggestion
			Help: "The total number of Fingerprint requests",
```

internal/services/fingerprint/consumer.go
```diff
@@ -148,6 +150,8 @@ func (c *Consumer) HandleDeviceFingerprint(ctx context.Context, event *Event) er
 			c.logger.Error().Msgf("could not marshal userdevice metadata for device: %s", ad.R.VehicleToken.R.UserDevice.ID)
 			return err
 		}
+
+		appmetrics.FingerprintRequestCount.With(prometheus.Labels{"protocol": *md.CANProtocol}).Inc()
```
JamesReate, line 154: there could be a case where this would nil pointer dereference. Consider using the `protocol` variable above instead (returned by ExtractProtocol)

internal/appmetrics/counters.go
```diff
@@ -70,4 +70,12 @@ var (
...
+		},
+		[]string{"protocol"},
```
JamesReate, line 79: consider having labels for success and failed states? ie. when return err in consumer.go could increment the request count but with a failure label

PR #229, opened 2023-11-21
SI-2176-preload-jsonld-vc-issuer-2

Allyson-English:
this proposes a way to keep schema.org included in our `@context`


elffjs:
I don't really know what this is doing. Is it mergeable?

PR #237, opened 2023-12-13
SI-2213-amd-addr-reset-event

Allyson-English:
When an AftermarketDeviceAddressReset event is received, update the 0x address of the device to reflect the change

### Notes:
- ethereum_address is a primary key in the aftermarket_devices table so this deletes the record and re-inserts it
- regenerated registry.go

### DIFF
--- a/internal/contracts/registry.go
+++ b/internal/contracts/registry.go
@@ -26,8 +26,13 @@ var (
 	_ = common.Big1
 	_ = types.BloomLookup
 	_ = event.NewSubscription
-	_ = abi.ConvertType
 )
+
+// AftermarketDeviceIdAddressPair is an auto generated low-level Go binding around an user-defined struct.
+type AftermarketDeviceIdAddressPair struct {
+	AftermarketDeviceNodeId *big.Int
+	DeviceAddress           common.Address
+}
 
 // AftermarketDeviceInfos is an auto generated low-level Go binding around an user-defined struct.
 type AftermarketDeviceInfos struct {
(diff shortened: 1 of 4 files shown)


internal/services/contracts_events_consumer.go
```diff
@@ -689,6 +692,35 @@ func (c *ContractsEventsConsumer) dcnNewExpiration(e *ContractEventData) error {
 	return nil
 }
 
+// aftermarketDeviceAddressReset handles the event of the same name from the registry contract.
+func (c *ContractsEventsConsumer) aftermarketDeviceAddressReset(e *ContractEventData) error {
+	if e.ChainID != c.settings.DIMORegistryChainID || e.Contract != common.HexToAddress(c.settings.DIMORegistryAddr) {
```
Allyson-English, line 697: I noticed a comment elsewhere in this file that it would be great to remove this from each handler. Could we move it up to the case/ switch and group all events that are from a common contract together so we could just check it once?

internal/services/contracts_events_consumer.go
```diff
@@ -689,6 +692,35 @@ func (c *ContractsEventsConsumer) dcnNewExpiration(e *ContractEventData) error {
...
+	amd, err := models.AftermarketDevices(
+		models.AftermarketDeviceWhere.TokenID.EQ(utils.BigToDecimal(args.TokenId)),
+	).One(context.Background(), c.db.DBS().Reader)
+	if err != nil {
+		return err
+	}
+
+	amd.EthereumAddress = args.AftermarketDeviceAddress.Bytes()
+
+	return amd.Upsert(
```
elffjs, line 715: Why an upsert? `amd` exists. If it doesn't, this will probably fail because `serial` is required.

internal/services/contracts_events_consumer.go
```diff
@@ -689,6 +692,35 @@ func (c *ContractsEventsConsumer) dcnNewExpiration(e *ContractEventData) error {
...
+		context.Background(),
+		c.db.DBS().Writer,
+		true,
+		[]string{models.AftermarketDeviceColumns.EthereumAddress},
+		boil.Columns{Cols: []string{models.AftermarketDeviceColumns.EthereumAddress}},
```
elffjs, line 720: I think you want `boil.Whitelist`. And I think you may have to whitelist `updated_at` as well.

internal/services/contracts_events_consumer.go
elffjs: We should do that, but don't worry about it now.

internal/contracts/registry.go
```diff
@@ -26,9 +26,14 @@ var (
 	_ = common.Big1
 	_ = types.BloomLookup
 	_ = event.NewSubscription
-	_ = abi.ConvertType
 )
 
+// AftermarketDeviceIdAddressPair is an auto generated low-level Go binding around an user-defined struct.
```
elffjs, line 31: Ah, you regenerated, that's good.

internal/services/contracts_events_consumer.go
elffjs: Let's add a test too.

internal/services/contracts_events_consumer.go
```diff
@@ -689,6 +694,29 @@ func (c *ContractsEventsConsumer) dcnNewExpiration(e *ContractEventData) error {
...
+	_, err := queries.Raw(fmt.Sprintf(
+		`UPDATE devices_api.%s 
+		SET ethereum_address = decode('%s', 'hex')
+		WHERE token_id = %d;`,
+		models.TableNames.AftermarketDevices,
+		strings.TrimPrefix(args.AftermarketDeviceAddress.String(), "0x"),
+		args.TokenId,
+	)).Exec(c.db.DBS().Writer)
+	return err
+
```
elffjs, line 717:
```suggestion
```

internal/services/contracts_events_consumer.go
```diff
@@ -689,6 +694,28 @@ func (c *ContractsEventsConsumer) dcnNewExpiration(e *ContractEventData) error {
...
+	var args contracts.RegistryAftermarketDeviceAddressReset
+	if err := json.Unmarshal(e.Arguments, &args); err != nil {
+		return err
+	}
+
+	_, err := queries.Raw(fmt.Sprintf(
+		`UPDATE devices_api.%s 
+		SET ethereum_address = decode('%s', 'hex')
+		WHERE token_id = %d;`,
```
elffjs, line 711: Could we use `?` or `$1` or something for these? Is that allowed?

elffjs: I always have to check that `decode` is being used correctly. I think it is.

internal/services/contracts_events_consumer.go
```diff
@@ -689,6 +693,30 @@ func (c *ContractsEventsConsumer) dcnNewExpiration(e *ContractEventData) error {
 	return nil
 }
 
+// $1 is the updated aftermarket device address, $2 is the token id
+const aftermarketDeviceAddressResetQuery = `
```
elffjs, line 697: I'm kinda happier to have this inline. Is that picky?

elffjs (approved):

PR #262, opened 2024-02-02
Remove watermill and replace with shared consumer event handling

0xdev22:
# Proposed Changes


cmd/devices-api/main.go
```diff
@@ -203,7 +203,7 @@ func startTaskStatusConsumer(logger zerolog.Logger, settings *config.Settings, p
 }
 
 func startContractEventsConsumer(logger zerolog.Logger, settings *config.Settings, pdb db.Store, autoPi *autopi.Integration, macaron *macaron.Integration, ddSvc services.DeviceDefinitionService) {
-	clusterConfig := sarama.NewConfig()
+	/*clusterConfig := sarama.NewConfig()
```
elffjs, line 206: Let's not commit a big comment. If it's going, it's going.

internal/services/contracts_events_consumer.go
```diff
@@ -102,34 +102,24 @@ func NewContractsEventsConsumer(pdb db.Store, log *zerolog.Logger, settings *con
...
-func (c *ContractsEventsConsumer) processMessage(msg *message.Message) error {
-	// Keep the pipeline moving no matter what.
-	defer func() { msg.Ack() }()
+func (c *ContractsEventsConsumer) RunConsumer() error {
+	ctx := context.Background()
 
-	// Deletion messages. We're the only actor that produces these, so ignore them.
-	if msg.Payload == nil {
-		return nil
+	if err := kafka.Consume[*shared.CloudEvent[json.RawMessage]](ctx, kafka.Config{
```
elffjs, line 108: Can we just return this error value to the caller? We're calling `.Fatal` here and in the caller, it seems duplicative.

cmd/devices-api/main.go
```diff
@@ -217,10 +217,13 @@ func startContractEventsConsumer(logger zerolog.Logger, settings *config.Setting
...
 		logger.Fatal().Err(err).Msg("Could not start contract event consumer")
-	}
+	}*/
 
 	cevConsumer := services.NewContractsEventsConsumer(pdb, &logger, settings, autoPi, macaron, ddSvc)
-	consumer.Start(context.Background(), cevConsumer.ProcessContractsEventsMessages)
+	if err := cevConsumer.RunConsumer(); err != nil {
+		logger.Fatal().Err(err).Msg("error occurred processing contract events")
+	}
+	// consumer.Start(context.Background(), cevConsumer.ProcessContractsEventsMessages)
```
elffjs, line 226: Same here.

elffjs (changes requested):

elffjs:
Giving this a bump.

internal/services/contracts_events_consumer.go
elffjs: Do we need to provide this type parameter, by the way? I feel like Go can infer this.

elffjs (approved):

PR #264, opened 2024-02-15
Implement new tesla API

0xdev22:
# Proposed Changes

- Implement new tesla api
- Create api route for retrieving vehicles using new api
- Handle tesla auth 
- Write tests to validate new endpoint

### Impacted Routes

Creates a new route `v1/user/devices/:integrationID`


internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1856,63 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
+// GetUserDevicesByIntegration godoc
+// @Description Get user devices using integration id
+// @Tags        user-devices
+// @Produce     json
+// @Accept      json
+// @Param       user_device body controllers.GetUserDevicesByIntegration true "all fields are required"
+// @Security    ApiKeyAuth
+// @Success     200 {object} controllers.GetUserDevicesByIntegrationResponse
+// @Security    BearerAuth
+// @Router      /user/devices/:integrationID [post]
```
elffjs, line 1868: I don't think this URI makes sense. Integration id under devices?

internal/controllers/user_devices_controller.go
```diff
@@ -1919,6 +1976,18 @@ type AdminRegisterUserDevice struct {
 	Verified    bool    `json:"verified"`
 }
 
+type GetUserDevicesByIntegration struct {
```
elffjs, line 1979: This name seems off. Maybe mention that it's a request?

internal/services/tesla_service.go
```diff
@@ -13,6 +15,8 @@ import (
 type TeslaService interface {
 	GetVehicle(ownerAccessToken string, id int) (*TeslaVehicle, error)
 	WakeUpVehicle(ownerAccessToken string, id int) error
+	CompleteTeslaAuthCodeExchange(authCode, redirectURI, region string) (*TeslaAuthCodeResponse, error)
+	GetVehicles(token, region string) ([]TeslaVehicle, error)
```
elffjs, line 19: I would just make a different service. This is a new API.

internal/services/tesla_service.go
```diff
@@ -82,3 +86,101 @@ type TeslaVehicle struct {
 	VehicleID int    `json:"vehicle_id"`
 	VIN       string `json:"vin"`
 }
+
+type GetVehiclesResponse struct {
+	Response *[]TeslaVehicle `json:"response"`
```
elffjs, line 91: I don't understand the pointer here.

internal/services/tesla_service.go
```diff
@@ -82,3 +86,101 @@ type TeslaVehicle struct {
...
+	AccessToken  string `json:"access_token"`
+	RefreshToken string `json:"refresh_token"`
+	IDToken      string `json:"id_token"`
+	ExpiresIn    int    `json:"expires_in"`
+	State        string `json:"state"`
+	TokenType    string `json:"token_type"`
+}
+
+func (t *teslaService) CompleteTeslaAuthCodeExchange(authCode, redirectURI, region string) (*TeslaAuthCodeResponse, error) {
+	u, _ := url.Parse("https://auth.tesla.com/oauth2/v3/token")
```
elffjs, line 112: I rewrote this using the `oauth2` package, if that's cleaner

https://github.com/DIMO-Network/testla/blob/main/auth.go

internal/services/tesla_service.go
```diff
@@ -82,3 +86,101 @@ type TeslaVehicle struct {
...
+	f := url.Values{}
+	f.Set("grant_type", "authorization_code")
+	f.Set("client_id", t.Settings.TeslaAuthorization.ClientID)
+	f.Set("client_secret", t.Settings.TeslaAuthorization.ClientSecret)
+	f.Set("code", authCode)
+	f.Set("redirect_uri", redirectURI)
+	f.Set("scope", "openid offline_access "+strings.Join(teslaScopes, " "))
+	f.Set("audience", fmt.Sprintf("https://fleet-api.prd.%s.vn.cloud.tesla.com", region))
```
elffjs, line 120: Can we validate that region is one of the two we want?

internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1856,63 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
+		if err != nil {
+			return err // swap this for http response
+		}
+
+		vehicles, err := udc.teslaService.GetVehicles(teslaAuth.AccessToken, reqBody.Region)
+		if err != nil {
+			return err
+		}
+
+		response := []GetUserDevicesByIntegrationResponse{}
```
elffjs, line 1900: Can we wrap the array in a wrapper struct?

internal/services/tesla_service.go
KevinJoiner, line 112: We should handle this error case.
```suggestion
	u, err := url.Parse("https://auth.tesla.com/oauth2/v3/token")
	if err != nil{
	   return nil, fmt.Errorf("failed to parse telsa URL: %w", err)
	}
```

internal/services/tesla_service.go
```diff
@@ -82,3 +86,101 @@ type TeslaVehicle struct {
...
+
+	req, err := http.NewRequest("POST", u.String(), strings.NewReader(f.Encode()))
```
KevinJoiner, line 122: We May want to use a context here so this request doesn't hang our program.
Ideally, this function would take in a context, and we can use that directly.
we can also use the http.MethodPost constant.
```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30) 
	defer cancel()
	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), strings.NewReader(f.Encode()))
```

internal/services/tesla_service.go
```diff
@@ -82,3 +86,101 @@ type TeslaVehicle struct {
...
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		errBody := new(TeslaError)
+		if err := json.NewDecoder(res.Body).Decode(errBody); err != nil {
+			return nil, err
```
KevinJoiner, line 137: If we hit this state, the caller should know what happened. As it is this will look like a random decoded error.
```suggestion
			return nil, fmt.Errorf("error occurred completing authorization, body could not be decoded: %w", err)
```

internal/services/tesla_service.go
```diff
@@ -82,3 +86,101 @@ type TeslaVehicle struct {
...
+	return teslaAuth, nil
+}
+
+func (t *teslaService) GetVehicles(token, region string) ([]TeslaVehicle, error) {
+	u := &url.URL{
+		Scheme: "https",
+		Host:   fmt.Sprintf("fleet-api.prd.%s.vn.cloud.tesla.com", region),
+		Path:   "api/1/vehicles",
+	}
+	req, err := http.NewRequest("GET", u.String(), nil)
```
KevinJoiner, line 156: Same as above for using a context
```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
	defer cancel()
	req, err : = http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
```

internal/services/tesla_service.go
```diff
@@ -82,3 +86,101 @@ type TeslaVehicle struct {
...
+	if err != nil {
+		return []TeslaVehicle{}, err
+	}
+
+	req.Header.Set("Authorization", "Bearer "+token)
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return []TeslaVehicle{}, err
```
KevinJoiner, line 164: If we have an error here or below, we should return the zero value of nil
```suggestion
		return nil, err
```

internal/services/tesla_service.go
```diff
@@ -82,3 +86,101 @@ type TeslaVehicle struct {
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		errBody := new(TeslaError)
+		if err := json.NewDecoder(resp.Body).Decode(errBody); err != nil {
+			return []TeslaVehicle{}, err
```
KevinJoiner, line 171: similar to above on giving the caller more info on what why they got a decode error.

internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1856,63 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
+		return fiber.NewError(fiber.StatusBadRequest, "Couldn't parse request JSON body.")
+	}
+
+	logger := udc.log.With().
+		Str("integrationId", integrationID).
+		Str("route", c.Route().Path).
+		Logger()
+	logger.Info().Msg("Attempting to fetch devices by integrationID")
+
+	deviceInteg, err := udc.DeviceDefSvc.GetIntegrationByID(ctx, integrationID)
```
KevinJoiner, line 1883: We should have some safeguards against long request
```suggestion
	reqCTX, cancel := context.WithTimeout(ctx, time.Second*30)
	defer cancel()
	deviceInteg, err := udc.DeviceDefSvc.GetIntegrationByID(reqCTX, integrationID)
```

internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1856,63 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
+// @Tags        user-devices
+// @Produce     json
+// @Accept      json
+// @Param       user_device body controllers.GetUserDevicesByIntegration true "all fields are required"
+// @Security    ApiKeyAuth
+// @Success     200 {object} controllers.GetUserDevicesByIntegrationResponse
+// @Security    BearerAuth
+// @Router      /user/devices/:integrationID [post]
+func (udc *UserDevicesController) GetUserDevicesByIntegration(c *fiber.Ctx) error {
+	ctx := context.Background()
```
KevinJoiner, line 1870: We can use the context of the request So we can clean up if the HTTP request gets canceled.
```suggestion
	ctx := c.Context()
```

internal/controllers/user_devices_controller.go
elffjs: I think it would be nice to give them MMY and maybe device definition id.

internal/controllers/user_devices_controller.go
```diff
@@ -1919,6 +1976,18 @@ type AdminRegisterUserDevice struct {
...
 
+type GetUserDevicesByIntegration struct {
+	AuthorizationCode string `json:"authorizationCode"`
+	RedirectURI       string `json:"redirectUri"`
+	Region            string `json:"region"`
+}
+
+type GetUserDevicesByIntegrationResponse struct {
+	ID        int    `json:"id"`
+	VehicleID int    `json:"vehicle_id"`
```
KevinJoiner, line 1987: I think we decided to stick with camelCase
```suggestion
	VehicleID int    `json:"vehicleID"`
```

internal/controllers/user_devices_controller_test.go
```diff
@@ -1012,6 +1015,53 @@ func (s *UserDevicesControllerTestSuite) TestPostRefreshSmartCarRateLimited() {
...
+	s.teslaSvc.EXPECT().GetVehicles(mockAuthCodeResp.AccessToken, mockRegion).Return(resp, nil)
+
+	request := test.BuildRequest("POST", "/user/devices/"+constants.TeslaVendor, fmt.Sprintf(`{
+		"authorizationCode": "%s",
+		"redirectUri": "%s",
+		"region": "na"
+	}`, mockAuthCode, mockRedirectURI))
+	response, _ := s.app.Test(request)
+
+	assert.Equal(s.T(), fiber.StatusOK, response.StatusCode)
```
KevinJoiner, line 1057: fun fact: suites have an `Assert` and `Require` function. Both version behave completely the same so there is no reason to change what you have here just sharing.
```suggestion
	s.Assert().Equal(fiber.StatusOK, response.StatusCode)
```

internal/controllers/user_devices_controller.go
```diff
@@ -1919,6 +1976,18 @@ type AdminRegisterUserDevice struct {
 	Verified    bool    `json:"verified"`
 }
 
+type GetUserDevicesByIntegration struct {
+	AuthorizationCode string `json:"authorizationCode"`
+	RedirectURI       string `json:"redirectUri"`
```
KevinJoiner, line 1981:
```suggestion
	RedirectURI       string `json:"redirectURI"`
```

internal/services/tesla_service.go
```diff
@@ -82,3 +86,101 @@ type TeslaVehicle struct {
...
 }
+
+type GetVehiclesResponse struct {
+	Response *[]TeslaVehicle `json:"response"`
+}
+
+type TeslaError struct {
+	Error            string
+	ErrorDescription string `json:"error_description"`
+	ReferenceID      string
```
KevinJoiner, line 97: might be worth adding tags to `Error` and `ReferenceID` for completeness.

KevinJoiner (changes requested):
Most of my comments are related to code maintainability. The general logic here makes sense to me.

cmd/devices-api/api.go
```diff
@@ -232,6 +232,7 @@ func startWebAPI(logger zerolog.Logger, settings *config.Settings, pdb db.Store,
 	v1Auth.Post("/user/devices/fromvin", userDeviceController.RegisterDeviceForUserFromVIN)
 	v1Auth.Post("/user/devices/fromsmartcar", userDeviceController.RegisterDeviceForUserFromSmartcar)
 	v1Auth.Post("/user/devices", userDeviceController.RegisterDeviceForUser)
+	v1Auth.Post("/user/devices/:integrationID", userDeviceController.GetUserDevicesByIntegration)
```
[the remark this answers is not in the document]
0xdev22: The body being passed to the API contains fields `authorizationCode` which I don't think we should be passing using `GET`

internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1856,63 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
+				ID:        v.ID,
+				VehicleID: v.VehicleID,
+				VIN:       v.VIN,
+			})
+		}
+		return c.JSON(response)
+	default:
+		logger.Error().Str("vendor", vendor).Msg("Attempted to fetch devices by integration")
+		return fiber.NewError(fiber.StatusBadRequest, fmt.Sprintf("unsupported integration %s", integrationID))
+	}
```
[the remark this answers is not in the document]
0xdev22: I will be removing the switch and vendor check entitrely

internal/services/tesla_service.go
0xdev22: You have a point, however this most likely will be an error from a 3rd party and `body could not be decoded` doesn't quite give a clue as to what happened which could be a temporary issue or a more permanent one, plus there is little to nothing the caller can do about this particular error. 

Maybe a better way to handle this is 
- Log the error on the server using a local logger
- change the error message for client to `error occurred completing authorization, invalid response received from during authorization process`

Good observation

internal/controllers/user_devices_controller.go
0xdev22: `redirectURI` is more a Go covention as opposed to what a client like say JS uses

cmd/devices-api/api.go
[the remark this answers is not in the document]
elffjs: I think it's confusing because the handler started with "Get". Let's align the terminology all the way down.

internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1856,63 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
+	deviceInteg, err := udc.DeviceDefSvc.GetIntegrationByID(ctx, integrationID)
+	if err != nil {
+		return shared.GrpcErrorToFiber(err, "failed to get integration by id")
+	}
+
+	switch vendor := deviceInteg.Vendor; vendor {
+	case constants.TeslaVendor:
+		teslaAuth, err := udc.teslaService.CompleteTeslaAuthCodeExchange(reqBody.AuthorizationCode, reqBody.RedirectURI, reqBody.Region)
+		if err != nil {
+			return err // swap this for http response
```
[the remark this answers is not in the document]
elffjs: I agree with wrapping this one for sure.

If it's some freak thing like the database connection failing I get a little lazy, but definitely this.

internal/controllers/user_devices_controller.go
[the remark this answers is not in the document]
elffjs: Well, we should check the vendor. I think that's good to do. This endpoint is not going to work for Smartcar now.

KevinJoiner (changes requested):

internal/services/tesla_service.go
elffjs: I'm tempted to let this slide, but I don't think it'll hurt.

If we use the `oauth2` library this logic shifts around a bit.

internal/services/tesla_service.go
elffjs: Definitely let's do this.

internal/controllers/user_devices_controller.go
elffjs: `vehicleId`, I would think.

internal/controllers/user_devices_controller.go
elffjs: I would do `redirectUri`.

cmd/devices-api/api.go
```diff
@@ -232,6 +233,7 @@ func startWebAPI(logger zerolog.Logger, settings *config.Settings, pdb db.Store,
 	v1Auth.Post("/user/devices/fromvin", userDeviceController.RegisterDeviceForUserFromVIN)
 	v1Auth.Post("/user/devices/fromsmartcar", userDeviceController.RegisterDeviceForUserFromSmartcar)
 	v1Auth.Post("/user/devices", userDeviceController.RegisterDeviceForUser)
+	v1Auth.Post("/user/devices/:tokenID/credentials", userDeviceController.GetUserDevicesWithCredentials)
```
elffjs, line 236: I think this is a dangerous path—`user/devices/{something}` is already taken and it makes it seem like the token id will be for a device. We're really acting on the integration. Can we do `/v1/integration/2/credentials`?

internal/config/settings.go
```diff
@@ -88,8 +88,15 @@ type Settings struct {
 
 	SyntheticFingerprintTopic         string `yaml:"SYNTHETIC_FINGERPRINT_TOPIC"`
 	SyntheticFingerprintConsumerGroup string `yaml:"SYNTHETIC_FINGERPRINT_CONSUMER_GROUP"`
+	TeslaAuthorization                Tesla  `yaml:"TESLA_AUTHORIZATION"`
```
elffjs, line 91: Maybe just call it `TESLA`?

internal/controllers/device_data_controller_test.go
```diff
@@ -147,7 +147,7 @@ func TestUserDevicesController_GetUserDeviceStatus(t *testing.T) {
 	}()
 
 	testUserID := "123123"
-	c := NewUserDevicesController(&config.Settings{Port: "3000"}, pdb.DBS, &mockDeps.logger, mockDeps.deviceDefSvc, mockDeps.deviceDefIntSvc, &fakeEventService{}, mockDeps.scClient, mockDeps.scTaskSvc, mockDeps.teslaSvc, mockDeps.teslaTaskService, nil, nil, mockDeps.nhtsaService, mockDeps.autoPiIngest, mockDeps.deviceDefinitionIngest, mockDeps.autoPiTaskSvc, nil, nil, nil, nil, mockDeps.openAISvc, nil, mockDeps.deviceDataSvc, nil, nil, nil, mockDeps.valuationsSvc)
+	c := NewUserDevicesController(&config.Settings{Port: "3000"}, pdb.DBS, &mockDeps.logger, mockDeps.deviceDefSvc, mockDeps.deviceDefIntSvc, &fakeEventService{}, mockDeps.scClient, mockDeps.scTaskSvc, mockDeps.teslaSvc, mockDeps.teslaTaskService, nil, nil, mockDeps.nhtsaService, mockDeps.autoPiIngest, mockDeps.deviceDefinitionIngest, mockDeps.autoPiTaskSvc, nil, nil, nil, nil, mockDeps.openAISvc, nil, mockDeps.deviceDataSvc, nil, nil, nil, mockDeps.valuationsSvc, nil)
```
elffjs, line 150: Can we just make this a new controller? I feel weird adding to this massive list.

internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1859,60 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
+		return fiber.NewError(fiber.StatusBadRequest, "Couldn't parse request JSON body.")
+	}
+
+	if reqBody.Region != "na" && reqBody.Region != "eu" {
+		return fiber.NewError(fiber.StatusBadRequest, "invalid value provided for region, only na and eu are allowed")
+	}
+
+	logger := udc.log.With().
+		Str("region", reqBody.Region).
+		Str("redirectURI", reqBody.RedirectURI).
```
elffjs, line 1884: Nitpick: I've been trying to do JS-type casing on the field names for logs.

internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1859,60 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
+		Str("route", c.Route().Path).
+		Logger()
+	logger.Info().Msg("Attempting to complete Tesla authorization")
+
+	teslaAuth, err := udc.teslaAPIService.CompleteTeslaAuthCodeExchange(reqBody.AuthorizationCode, reqBody.RedirectURI, reqBody.Region)
```
elffjs, line 1889: Again, I would like to separate out the new "V2" Fleet API stuff. It's separate. Make a new client/service, I think.

internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1859,60 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
+
+	vehicles, err := udc.teslaAPIService.GetVehicles(teslaAuth.AccessToken, reqBody.Region)
+	if err != nil {
+		return err
+	}
+
+	response := make([]GetUserDevicesWithCredentialsResponse, 0, len(vehicles))
+
+	for _, v := range vehicles {
+		response = append(response, GetUserDevicesWithCredentialsResponse{
```
elffjs, line 1902: Make `response[i] = ...`?

internal/controllers/user_devices_controller.go
```diff
@@ -1919,6 +1976,26 @@ type AdminRegisterUserDevice struct {
...
+}
+
+type GetUserDevicesWithCredentialsResponse struct {
+	ID                 int    `json:"id"`
+	VehicleID          int    `json:"vehicleId"`
+	VIN                string `json:"vin"`
+	Make               string
+	Model              string
+	Year               int
+	DeviceDefinitionID string
```
elffjs, line 1996: Are we populating these?

We should probably nest the device definition stuff, so like

```
{
  vehicles: [
    {id: "5434637", vin: "5YJ...", definition: {id: "2fU8988...", make: "Toyota", ....}},
   ...
  ]
}

File: internal/controllers/user_devices_controller.go [hunk 29]
```diff
@@ -1856,6 +1859,60 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
+	vehicles, err := udc.teslaAPIService.GetVehicles(teslaAuth.AccessToken, reqBody.Region)
+	if err != nil {
+		return err
+	}
+
+	response := make([]GetUserDevicesWithCredentialsResponse, 0, len(vehicles))
+
+	for _, v := range vehicles {
+		response = append(response, GetUserDevicesWithCredentialsResponse{
+			ID:        v.ID,
```
elffjs, line 1903: Can we make this a string? I feel like other integrations might have non-numeric ids; with a string we can respond to all of that.

internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1859,60 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
+// GetUserDevicesWithCredentials godoc
+// @Description Complete Tesla auth and get devices for authenticated user
+// @Tags        user-devices
+// @Produce     json
+// @Accept      json
+// @Param       user_device body controllers.GetUserDevicesByIntegrationRequest true "all fields are required"
+// @Security    ApiKeyAuth
+// @Success     200 {object} controllers.GetUserDevicesByIntegrationResponse
+// @Security    BearerAuth
+// @Router      /user/devices/{tokenID}/credentials [post]
```
elffjs, line 1871: Do we ever use `tokenID` here? I feel like something that says "I'm trying to connect to Tesla" should be in the path.

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/DIMO-Network/devices-api/internal/config"
+	"github.com/rs/zerolog"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+)
+
+//go:generate mockgen -source tesla_api_service.go -destination mocks/tesla_api_service_mock.go
+type TeslaAPIService interface {
```
elffjs, line 16: Can we put the word "fleet API" in here somewhere?

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,129 @@
...
+		Settings: settings,
+		HTTPClient: &http.Client{
+			Timeout: 5 * time.Second,
+		},
+		log: logger,
+	}
+}
+
+func (t *teslaAPIService) CompleteTeslaAuthCodeExchange(authCode, redirectURI, region string) (*TeslaAuthCodeResponse, error) {
+	u, err := url.Parse(t.Settings.TeslaAuthorization.TokenAuthURL)
```
elffjs, line 38: Did you see the oauth2 code I posted? Does that read better to you at all?

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,129 @@
...
+	f := url.Values{}
+	f.Set("grant_type", "authorization_code")
+	f.Set("client_id", t.Settings.TeslaAuthorization.ClientID)
+	f.Set("client_secret", t.Settings.TeslaAuthorization.ClientSecret)
+	f.Set("code", authCode)
+	f.Set("redirect_uri", redirectURI)
+	f.Set("scope", "openid offline_access "+strings.Join(teslaScopes, " "))
+	f.Set("audience", fmt.Sprintf("https://fleet-api.prd.%s.vn.cloud.tesla.com", region))
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
```
elffjs, line 51: Can someone remind me how this interacts with the 5 second timeout that's on the HTTP client we're using? Do we need both?

Note that it's still good to pass in the context so that canceling the request can kill this request to Tesla.

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,129 @@
...
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), strings.NewReader(f.Encode()))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+
+	res, err := http.DefaultClient.Do(req)
```
elffjs, line 60: Oh, but we're using the default client here. What about the client that's on `teslaAPIService`? Let's use that.

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,129 @@
...
+		return nil, err
+	}
+
+	return teslaAuth, nil
+}
+
+func (t *teslaAPIService) GetVehicles(token, region string) ([]TeslaVehicle, error) {
+	u := &url.URL{
+		Scheme: "https",
+		Host:   fmt.Sprintf("fleet-api.prd.%s.vn.cloud.tesla.com", region),
```
elffjs, line 89: Can we parametrize this as a setting as well? I'm trying to envision a world where we're mocking this stuff locally.

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,129 @@
...
+		return nil, fmt.Errorf("error occurred completing authorization: %s", errBody.ErrorDescription)
+	}
+
+	vehicles := new(GetVehiclesResponse)
+	if err := json.NewDecoder(resp.Body).Decode(vehicles); err != nil {
+		return nil, err
+	}
+
+	if vehicles.Response == nil {
+		return nil, fmt.Errorf("Error occurred fetching vehicles")
```
elffjs, line 125: I'm surprised it's not complaining about the capital letter here.

elffjs (changes requested):

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,129 @@
...
+	return &teslaAPIService{
+		Settings: settings,
+		HTTPClient: &http.Client{
+			Timeout: 5 * time.Second,
+		},
+		log: logger,
+	}
+}
+
+func (t *teslaAPIService) CompleteTeslaAuthCodeExchange(authCode, redirectURI, region string) (*TeslaAuthCodeResponse, error) {
```
elffjs, line 37: Could we do quick docstrings on these too? Sorry.

internal/services/tesla_api_service.go
elffjs: Let me remind

https://github.com/DIMO-Network/testla/blob/799589ea9829c2aed5aedc741aad38575f7c98a3/auth.go#L49

internal/services/tesla_api_service.go
KevinJoiner: The Timout set on the client will override the one used in the context. Personally, I don't love using short Timeouts on clients because it is a hard cap on all requests and encompasses more than just sending the request since it will also close the body if you do not read it before the timeout is exceeded.

internal/controllers/user_devices_controller.go
KevinJoiner: For this remember to change `response := make([]GetUserDevicesWithCredentialsResponse, 0, len(vehicles))` => `response := make([]GetUserDevicesWithCredentialsResponse, len(vehicles))` once compiled I believe both have the same performance since the underlying array is allocated to the correct size. I usually lean towards `response[i] = v` but that is just for stylistic reasons.

cmd/devices-api/api.go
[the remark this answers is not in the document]
KevinJoiner: Yeah, the name just confused me. POST makes sense for sensitive data.

internal/controllers/user_devices_controller.go
```diff
@@ -1919,6 +1994,30 @@ type AdminRegisterUserDevice struct {
 	Verified    bool    `json:"verified"`
 }
 
+type CompleteOAuthExchangeRequest struct {
```
elffjs, line 1997: Can we add documentation strings and examples (see elsewhere) for these fields? Those will show up in the Swagger docs. I think that's a great help.

internal/controllers/user_devices_controller.go
elffjs: @KevinJoiner Does that make sense? I must admit I'm not super confident on this.

internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1859,78 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
 }
 
+// CompleteOAuthExchange godoc
+// @Description Complete Tesla auth and get devices for authenticated user
+// @Tags        user-devices
+// @Produce     json
+// @Accept      json
+// @Param       user_device body controllers.CompleteOAuthExchangeRequest true "all fields are required"
+// @Security    ApiKeyAuth
+// @Success     200 {object} controllers.CompleteOAuthExchangeResponse
```
elffjs, line 1869: Didn't we change this? There's a wrapper now.

internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1859,78 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
 	}, sigBytes)
 }
 
+// CompleteOAuthExchange godoc
+// @Description Complete Tesla auth and get devices for authenticated user
+// @Tags        user-devices
+// @Produce     json
+// @Accept      json
+// @Param       user_device body controllers.CompleteOAuthExchangeRequest true "all fields are required"
```
elffjs, line 1867: Don't we have a path param?

internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1859,78 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
+// @Security    ApiKeyAuth
+// @Success     200 {object} controllers.CompleteOAuthExchangeResponse
+// @Security    BearerAuth
+// @Router      /integration/:tokenID/credentials [post]
+func (udc *UserDevicesController) CompleteOAuthExchange(c *fiber.Ctx) error {
```
elffjs, line 1872: Let's check that the path param is really indicating Tesla.

internal/controllers/user_devices_controller.go
```diff
@@ -1919,6 +1994,30 @@ type AdminRegisterUserDevice struct {
...
+
+type CompleteOAuthExchangeResponseWrapper struct {
+	Vehicles []CompleteOAuthExchangeResponse
+}
+
+type DeviceDefinition struct {
+	Make               string `json:"make"`
+	Model              string `json:"model"`
+	Year               int    `json:"year"`
+	DeviceDefinitionID string `json:"deviceDefinitionId"`
```
elffjs, line 2011: Maybe just make this `id`.

internal/controllers/user_devices_controller.go
```diff
@@ -1919,6 +1994,30 @@ type AdminRegisterUserDevice struct {
...
+}
+
+type CompleteOAuthExchangeResponse struct {
+	ExternalID string `json:"id"`
```
elffjs, line 2015: Can we call it `externalId` in the JSON too?

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,130 @@
...
+	"golang.org/x/oauth2"
+)
+
+//go:generate mockgen -source tesla_api_service.go -destination mocks/tesla_api_service_mock.go
+type TeslaAPIService interface {
+	CompleteTeslaAuthCodeExchange(authCode, redirectURI, region string) (*TeslaAuthCodeResponse, error)
+	GetVehicles(token, region string) ([]TeslaVehicle, error)
+}
+
+var teslaScopes = []string{"openid offline_access", "user_data", "vehicle_device_data", "vehicle_cmds", "vehicle_charging_cmds", "energy_device_data", "energy_device_data", "energy_cmds"}
```
elffjs, line 23:
```suggestion
var teslaScopes = []string{"openid", "offline_access", "user_data", "vehicle_device_data", "vehicle_cmds", "vehicle_charging_cmds", "energy_device_data", "energy_device_data", "energy_cmds"}
```
Just to be consistent.

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,130 @@
...
+			TokenURL: t.Settings.Tesla.TokenURL,
+		},
+		RedirectURL: redirectURI,
+		Scopes:      teslaScopes,
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+
+	tok, err := conf.Exchange(ctx, authCode, oauth2.SetAuthURLParam("audience", fmt.Sprintf("https://fleet-api.prd.%s.vn.cloud.tesla.com", region)), oauth2.SetAuthURLParam("grant_type", "authorization_code"))
```
elffjs, line 72: You don't need to set `grant_type`, I don't think.

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		var e *oauth2.RetrieveError
+		errors.As(err, &e)
```
elffjs: I think we need more code here. `As` has a boolean return type.

https://pkg.go.dev/errors#example-As

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil, fmt.Errorf("error occurred completing authorization: %s", e.ErrorDescription)
+	}
+
+	return &TeslaAuthCodeResponse{
+		AccessToken:  tok.AccessToken,
+		RefreshToken: tok.RefreshToken,
+		ExpiresIn:    int(tok.Expiry.Unix()),
```
elffjs: I don't think this is right. `tok.Expiry` is the time at which this thing expires. It's not `expires_in`. It's `time.Now() + expires_in`.

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,130 @@
...
+type TeslaError struct {
+	Error            string `json:"error"`
+	ErrorDescription string `json:"error_description"`
+	ReferenceID      string `json:"ReferenceID"`
+}
+
+type TeslaAuthCodeResponse struct {
+	AccessToken  string `json:"access_token"`
+	RefreshToken string `json:"refresh_token"`
+	IDToken      string `json:"id_token"`
```
elffjs: Are we using this?

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,130 @@
...
+	ExpiresIn    int    `json:"expires_in"`
+	State        string `json:"state"`
+	TokenType    string `json:"token_type"`
```
elffjs: Are we using these two?

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,130 @@
...
+var teslaScopes = []string{"openid offline_access", "user_data", "vehicle_device_data", "vehicle_cmds", "vehicle_charging_cmds", "energy_device_data", "energy_device_data", "energy_cmds"}
+
+type GetVehiclesResponse struct {
+	Response []TeslaVehicle `json:"response"`
+}
+
+type TeslaError struct {
+	Error            string `json:"error"`
+	ErrorDescription string `json:"error_description"`
+	ReferenceID      string `json:"ReferenceID"`
```
elffjs: Weird, they switch from snake case to camel?

internal/controllers/user_integrations_auth_controller.go
```diff
@@ -0,0 +1,135 @@
...
+	teslaFleetAPISvc services.TeslaFleetAPIService,
+) UserIntegrationAuthController {
+	return UserIntegrationAuthController{
+		Settings:         settings,
+		DBS:              dbs,
+		DeviceDefSvc:     ddSvc,
+		log:              logger,
+		cipher:           cipher,
+		producer:         producer,
+		redisCache:       cache,
```
elffjs, line 43: Why do we have these three dependencies?

internal/controllers/user_devices_controller.go
elffjs: Should we move this to the file that uses it?

cmd/devices-api/api.go
```diff
@@ -357,9 +359,9 @@ func startWebAPI(logger zerolog.Logger, settings *config.Settings, pdb db.Store,
 		logger.Fatal().Err(err).Msg("Failed to create issuer.")
 	}
 
-	if err := fingerprint.RunConsumer(ctx, settings, &logger, iss, pdb); err != nil {
+	/*if err := fingerprint.RunConsumer(ctx, settings, &logger, iss, pdb); err != nil {
```
elffjs, line 362: Watch out!

internal/services/tesla_api_service.go
```diff
@@ -0,0 +1,129 @@
...
+type TeslaAPIService interface {
+	CompleteTeslaAuthCodeExchange(authCode, redirectURI, region string) (*TeslaAuthCodeResponse, error)
+	GetVehicles(token, region string) ([]TeslaVehicle, error)
+}
+
+type teslaAPIService struct {
+	Settings   *config.Settings
+	HTTPClient *http.Client
+	log        *zerolog.Logger
+}
```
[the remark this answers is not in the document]
elffjs: I think the suggestion is not that the interface is bad, more that the interface is closer to the consumer (the controller) than the implementer (the service). If the service adds a third method that the controller doesn't need, the controller doesn't care.

internal/services/tesla_fleet_api_service.go
```diff
@@ -0,0 +1,141 @@
...
+
+	"github.com/DIMO-Network/devices-api/internal/config"
+	"github.com/rs/zerolog"
+	"golang.org/x/oauth2"
+)
+
+//go:generate mockgen -source tesla_fleet_api_service.go -destination mocks/tesla_fleet_api_service_mock.go
+type TeslaFleetAPIService interface {
+	CompleteTeslaAuthCodeExchange(authCode, redirectURI, region string) (*TeslaAuthCodeResponse, error)
+	GetVehicles(token, region string) ([]TeslaVehicle, error)
```
elffjs, line 20: Let's add a context to each one of these calls.

internal/services/tesla_api_service.go
0xdev22: image

Not sure I follow your meaning, but this is what the API response from Tesla looks like

0xdev22: I'm not sure I understand what you mean. Also upon checking again, seems tesla returns expires_in and not expiry, but the oauth library seems to have expiry instead.

internal/services/tesla_api_service.go
0xdev22: Just included them for completeness since the API returns them, but I will remove.

internal/services/tesla_fleet_api_service.go
0xdev22: We have a context for timeout within both functions as well.

internal/services/tesla_api_service.go
elffjs: I think the `oauth2` library computes `Expiry` as `time.Now() + expires_in`, where `expires_in` is coming from the token.

internal/services/tesla_api_service.go
elffjs: But you're not populating them, are you? If you were populating them it would be one thing.

internal/controllers/user_devices_controller_test.go
```diff
@@ -75,6 +75,7 @@ type UserDevicesControllerTestSuite struct {
 	natsServer      *server.Server
 	userDeviceSvc   *mock_services.MockUserDeviceService
 	valuationsSrvc  *mock_services.MockValuationsAPIService
+	teslaSvc        *mock_services.MockTeslaService
```
elffjs, line 78: !

internal/controllers/user_integrations_auth_controller.go
```diff
@@ -0,0 +1,136 @@
...
+	VIN        string           `json:"vin"`
+	Definition DeviceDefinition `json:"definition"`
+}
+
+// CompleteOAuthExchange godoc
+// @Description Complete Tesla auth and get devices for authenticated user
+// @Tags        user-devices
+// @Produce     json
+// @Accept      json
+// @Param       tokenID path string                   true "token id"
```
elffjs, line 71: Can we say that it's the token id for the integration? For completeness.

internal/services/tesla_api_service.go
0xdev22: I'm populating tokenType, https://github.com/DIMO-Network/devices-api/pull/264/files/549f26e83f4b6488cd3a65ca0f88e54b0658b6fa#diff-e013fbe8446d2ccf3a214dda0a4f112e0e4850c595ddb2f43197e0d613a14973R83

The library doesn't seem to provide state

internal/services/tesla_fleet_api_service.go
elffjs: That's also good, but I think ideally the incoming request being killed should also kill the child requests going out.

internal/services/tesla_fleet_api_service.go
```diff
@@ -0,0 +1,141 @@
...
+		TokenType:    tok.TokenType,
+	}, nil
+}
+
+// GetVehicles godoc
+// @Description Call Tesla Fleet API to get a list of vehicles using authorization token
+// @Param       token - authorization token to be used as bearer token
+// @Param       region - API region which is used to determine which fleet api to call
+// @Success     200 {object} []services.TeslaVehicle
+func (t *teslaFleetAPIService) GetVehicles(token, region string) ([]TeslaVehicle, error) {
```
elffjs, line 98: We have a lot of naked `return nil, err`s here. Could we add some context to them?

internal/services/tesla_fleet_api_service.go
```diff
@@ -0,0 +1,141 @@
...
+		ExpiresIn:    int(tok.Expiry.Unix()),
+		TokenType:    tok.TokenType,
+	}, nil
+}
+
+// GetVehicles godoc
+// @Description Call Tesla Fleet API to get a list of vehicles using authorization token
+// @Param       token - authorization token to be used as bearer token
+// @Param       region - API region which is used to determine which fleet api to call
+// @Success     200 {object} []services.TeslaVehicle
```
elffjs, line 97: These are docs for routes, which these are not—I wouldn't say that a _function_ had response code 200. Just say what the functions do.

internal/services/tesla_api_service.go
elffjs: Yeah, see https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.17.0:oauth2.go;l=225

internal/controllers/user_devices_controller.go
elffjs: Like, if I hit `/integration/1/credentials` then this should fail.

internal/controllers/user_integrations_auth_controller.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil || len(decodeVIN.DeviceDefinitionId) == 0 {
+			u.log.Err(err).Str("deviceDefinitionID", decodeVIN.DeviceDefinitionId).Msg("An error occurred fetching device definition using ID")
+			return fiber.NewError(fiber.StatusFailedDependency, "An error occurred completing tesla authorization")
+		}
+
+		response = append(response, CompleteOAuthExchangeResponse{
+			ExternalID: strconv.Itoa(v.ID),
+			VehicleID:  v.VehicleID,
+			VIN:        v.VIN,
+			Definition: DeviceDefinition{
```
elffjs, line 122: Oh you've got the definitions in here now too, very good!

internal/services/tesla_fleet_api_service.go
```diff
@@ -0,0 +1,141 @@
...
+		ClientID:     t.Settings.Tesla.ClientID,
+		ClientSecret: t.Settings.Tesla.ClientSecret,
+		Endpoint: oauth2.Endpoint{
+			TokenURL: t.Settings.Tesla.TokenAuthURL,
+		},
+		RedirectURL: redirectURI,
+		Scopes:      teslaScopes,
+	}
+
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), time.Second*30)
```
elffjs, line 73: We would pass the context in here.

internal/services/tesla_fleet_api_service.go
```diff
@@ -0,0 +1,141 @@
...
+// @Param       token - authorization token to be used as bearer token
+// @Param       region - API region which is used to determine which fleet api to call
+// @Returns     {array} []services.TeslaVehicle
+func (t *teslaFleetAPIService) GetVehicles(ctx context.Context, token, region string) ([]TeslaVehicle, error) {
+	u := &url.URL{
+		Scheme: "https",
+		Host:   fmt.Sprintf("fleet-api.prd.%s.vn.cloud.tesla.com", region),
+		Path:   "api/1/vehicles",
+	}
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), time.Second*10)
```
elffjs, line 104: And here.

internal/services/tesla_fleet_api_service.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, fmt.Errorf("could not fetch vehicles for user: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		errBody := new(TeslaFleetAPIError)
+		if err := json.NewDecoder(resp.Body).Decode(errBody); err != nil {
+			t.log.
+				Err(err).
+				Str("url", fmt.Sprintf(t.Settings.Tesla.FleetAPI, region)).
```
elffjs, line 124: Are we using this setting anywhere else?

internal/services/tesla_fleet_api_service.go
```diff
@@ -0,0 +1,143 @@
...
+		if errors.As(err, &e) {
+			errString = e.ErrorDescription
+		}
+		return nil, fmt.Errorf("error occurred completing authorization: %s", errString)
+	}
+
+	return &TeslaAuthCodeResponse{
+		AccessToken:  tok.AccessToken,
+		RefreshToken: tok.RefreshToken,
+		Expiry:       int(tok.Expiry.Unix()),
```
elffjs, line 90: I guess we'll see how this gets used. I'm skeptical that the unix timestamp is going to be useful over the `time.Time.`

internal/services/tesla_fleet_api_service.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	ctxTimeout, cancel := context.WithTimeout(ctx, time.Second*30)
+	defer cancel()
+
+	tok, err := conf.Exchange(ctxTimeout, authCode, oauth2.SetAuthURLParam("audience", fmt.Sprintf("https://fleet-api.prd.%s.vn.cloud.tesla.com", region)), oauth2.SetAuthURLParam("grant_type", "authorization_code"))
+	if err != nil {
+		var e *oauth2.RetrieveError
+		errString := ""
+		if errors.As(err, &e) {
```
elffjs, line 81: If the error is not a RetrieveError, we'd still want to log it, no? It doesn't seem like we will.

internal/config/settings.go
```diff
@@ -88,8 +88,16 @@ type Settings struct {
...
 
 func (s *Settings) IsProduction() bool {
 	return s.Environment == "prod" // this string is set in the helm chart values-prod.yaml
 }
+
+type Tesla struct {
+	ClientID     string `yaml:"CLIENT_ID"`
+	ClientSecret string `yaml:"CLIENT_SECRET"`
+	TokenAuthURL string `yaml:"TOKEN_AUTH_URL"`
+	FleetAPI     string `yaml:"FLEET_API"`
```
elffjs, line 102: Let's provide a correct default for this in `values.yaml`, then.

internal/services/tesla_fleet_api_service.go
```diff
@@ -0,0 +1,143 @@
...
+			TokenURL: t.Settings.Tesla.TokenAuthURL,
+		},
+		RedirectURL: redirectURI,
+		Scopes:      teslaScopes,
+	}
+
+	ctxTimeout, cancel := context.WithTimeout(ctx, time.Second*30)
+	defer cancel()
+
+	tok, err := conf.Exchange(ctxTimeout, authCode, oauth2.SetAuthURLParam("audience", fmt.Sprintf("https://fleet-api.prd.%s.vn.cloud.tesla.com", region)))
```
elffjs, line 77: Another place where we could use our setting.

internal/controllers/user_devices_controller.go
```diff
@@ -140,6 +141,7 @@ func NewUserDevicesController(settings *config.Settings,
 	wallet services.SyntheticWalletInstanceService,
 	userDeviceSvc services.UserDeviceService,
 	valuationsAPISrv services.ValuationsAPIService,
+	teslaAPIService services.TeslaAPIService,
```
elffjs: Can we remove this now?

internal/controllers/user_integrations_auth_controller.go
```diff
@@ -0,0 +1,151 @@
...
+// @Router      /integration/:tokenID/credentials [post]
+func (u *UserIntegrationAuthController) CompleteOAuthExchange(c *fiber.Ctx) error {
+	tokenID := c.Params("tokenID")
+
+	tkID, err := strconv.ParseUint(tokenID, 10, 64)
+	if err != nil {
+		return fiber.NewError(fiber.StatusBadRequest, "could not process the provided tokenId!")
+	}
+
+	intd, err := u.DeviceDefSvc.GetIntegrationByTokenID(c.Context(), tkID)
```
elffjs, line 89: I think the linter is complaining about good things.

internal/controllers/user_integrations_auth_controller.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/rs/zerolog"
+	"strconv"
+)
+
+type UserIntegrationAuthController struct {
+	Settings         *config.Settings
+	DBS              func() *db.ReaderWriter
+	DeviceDefSvc     services.DeviceDefinitionService
+	log              *zerolog.Logger
+	cipher           shared.Cipher
```
elffjs, line 21: Here, too.

internal/services/tesla_fleet_api_service.go
```diff
@@ -0,0 +1,143 @@
...
+		AccessToken:  tok.AccessToken,
+		RefreshToken: tok.RefreshToken,
+		Expiry:       tok.Expiry,
+		TokenType:    tok.TokenType,
+	}, nil
+}
+
+// GetVehicles godoc
+// @Description Call Tesla Fleet API to get a list of vehicles using authorization token
+// @Param       token - authorization token to be used as bearer token
```
elffjs, line 97: Again, I would avoid doing this Swagger thing for normal Go functions. But we can let it slide here.

internal/services/tesla_fleet_api_service.go
```diff
@@ -0,0 +1,143 @@
...
+// @Param       region - API region which is used to determine which fleet api to call
+// @Returns     {array} []services.TeslaVehicle
+func (t *teslaFleetAPIService) GetVehicles(ctx context.Context, token, region string) ([]TeslaVehicle, error) {
+	u := &url.URL{
+		Scheme: "https",
+		Host:   fmt.Sprintf(t.Settings.Tesla.FleetAPI, region),
```
elffjs, line 103: I don't think we can use this constant both here and in the audience field

internal/config/settings.go
```diff
@@ -88,8 +88,16 @@ type Settings struct {
...
 
 func (s *Settings) IsProduction() bool {
 	return s.Environment == "prod" // this string is set in the helm chart values-prod.yaml
 }
+
+type Tesla struct {
+	ClientID     string `yaml:"CLIENT_ID"`
+	ClientSecret string `yaml:"CLIENT_SECRET"`
+	TokenURL     string `yaml:"TOKEN_URL"`
+	FleetAPI     string `yaml:"FLEET_API"`
```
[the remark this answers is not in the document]
elffjs: I'll try to fix it.

elffjs (approved):

elffjs:
I would like to have waited for Kevin's review.

Given that we've merged this, can you test it on dev?

0xdev22:
> I would like to have waited for Kevin's review.
> 
> Given that we've merged this, can you test it on dev?

Oops! Sorry about that @KevinJoiner .. 

Sure testing it now

PR #267, opened 2024-02-22
Persist auth response from tesla to redis and write tests for new function

0xdev22:
# Proposed Changes

### DIFF
--- a/internal/controllers/user_integrations_auth_controller.go
+++ b/internal/controllers/user_integrations_auth_controller.go
@@ -1,13 +1,21 @@
 package controllers
 
 import (
+	"context"
+	"encoding/json"
+	"fmt"
 	"github.com/DIMO-Network/devices-api/internal/config"
 	"github.com/DIMO-Network/devices-api/internal/constants"
+	"github.com/DIMO-Network/devices-api/internal/controllers/helpers"
 	"github.com/DIMO-Network/devices-api/internal/services"
+	"github.com/DIMO-Network/shared"
+	pb "github.com/DIMO-Network/shared/api/users"
 	"github.com/DIMO-Network/shared/db"
+	"github.com/DIMO-Network/shared/redis"
 	"github.com/gofiber/fiber/v2"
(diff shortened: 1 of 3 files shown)


internal/controllers/user_integrations_auth_controller.go
```diff
@@ -73,8 +90,18 @@ type DeviceDefinition struct {
 // @Security    BearerAuth
 // @Router      /integration/:tokenID/credentials [post]
 func (u *UserIntegrationAuthController) CompleteOAuthExchange(c *fiber.Ctx) error {
-	tokenID := c.Params("tokenID")
+	//Get the current user
```
elffjs, line 93:
```suggestion
	// Get the current user
```

elffjs:
I agree with the linter—we're not passing enough arguments to the controller constructor anymore.

elffjs (changes requested):
I agree with the linter—we're not passing enough arguments to the controller constructor anymore.

elffjs:
I think we must patch up some of the tests from last time to account for the new stuff in the handler. They're failing locally for me.

elffjs:
(Other than that, the PR looks good)

elffjs (approved):

internal/controllers/user_integrations_auth_controller.go
```diff
@@ -1,13 +1,21 @@
 package controllers
 
 import (
+	"context"
+	"encoding/json"
+	"fmt"
```
KevinJoiner, line 6: nit: Should separate the std lib packages from other packages with a newline 
https://go.dev/wiki/CodeReviewComments#imports

KevinJoiner (approved):
One nit but LGTM

internal/controllers/user_integrations_auth_controller.go
elffjs: Is there a linter rule for this?

I think VS Code does this by default (whatever they're running on save) but I think Doyin uses GoLand and it's doing something else./

PR #269, opened 2024-02-26
Feature/SI-2449: Add Tesla Fleet API implementation to register Tesla Device implementation

0xdev22:
# Proposed Changes
- Retrieve Tesla v2 credentials from cache
- Add version flag to metadata
- Persist Tesla v2 credentials to db and push to kafka
- Encrypt Tesla credentials for storage in db

### Impacted Routes
`/user/devices/:userDeviceID/integrations/:integrationID`

#### Tesla Auth on Redis

image

#### Persist Tesla Cred 

image

#### Kafka Event 

image

#### API Call response

image


internal/controllers/user_integrations_controller.go
```diff
@@ -1797,6 +1797,8 @@ func (udc *UserDevicesController) registerSmartcarIntegration(c *fiber.Ctx, logg
 }
 
 func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zerolog.Logger, tx *sql.Tx, userDeviceID string, integ *ddgrpc.Integration, ud *models.UserDevice) error {
+	// Flag for which api version should be used
+	apiVersion := c.QueryInt("version", 1)
```
elffjs, line 1801: Maybe this should be part of the body? It seems a little odd to me to have query parameters and a JSON body.

My feeling is that we would drop this flag in pretty short order.

internal/controllers/user_integrations_controller.go
```diff
@@ -1810,6 +1812,20 @@ func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zero
...
 	}
 
+	if reqBody.ExternalID == "" {
+		return fiber.NewError(fiber.StatusBadRequest, "Missing externalID parameter")
+	}
+
+	if apiVersion == 2 { // If version is 2, we are using fleet api which, which has token stored in cache
+		deviceIntReq, err := udc.getTeslaAuthFromCache(c.Context(), ud.UserID)
+		if err != nil {
+			return fiber.NewError(fiber.StatusBadRequest, "Couldn't parse request body.")
```
elffjs, line 1822: It's not quite the case that we couldn't parse the request body, is it? We definitely parsed the request body. "Couldn't retrieve stored credentials."?

internal/controllers/user_integrations_controller.go
elffjs, line 1801: It seems a little odd to me to have query parameters and a JSON body.

But I'm not too concerned. My feeling is that we would drop this flag in pretty short order. If we are going this route, can we add the query parameter to the Swagger?

internal/controllers/user_integrations_controller.go
```diff
@@ -1933,6 +1950,49 @@ func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zero
...
+		return nil, errors.Wrap(err, "failed to parse tesla authorization token")
+	}
+
+	if teslaAuth == nil {
+		return nil, fmt.Errorf("could not find any tesla authorization for user")
+	}
+
+	return &RegisterDeviceIntegrationRequest{
+		AccessToken:  teslaAuth.AccessToken,
+		ExpiresIn:    int(teslaAuth.Expiry.Unix()),
```
elffjs, line 1991: We keep mixing up expiry and "expires in". The first is a time, the second is a duration.

internal/controllers/user_integrations_controller.go
```diff
@@ -1933,6 +1950,49 @@ func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zero
...
+
+	if user.EthereumAddress == nil {
+		return nil, fmt.Errorf("missing wallet address for user")
+	}
+
+	cacheKey := fmt.Sprintf(TeslaFleetAuthCacheKey, *user.EthereumAddress)
+	encTeslaAuth, err := udc.redisCache.Get(ctx, cacheKey).Result()
+	if err != nil {
+		udc.log.Err(err).Str("Cache Key", cacheKey).Msg("Error occurred retrieving tesla auth from cache using the key")
+		return nil, errors.Wrap(err, "an error occurred completing tesla device registration")
```
elffjs, line 1967: Can we be more specific?

I'm a `fmt.Errorf` guy, personally, but `errors.Wrap` is good too.

internal/controllers/user_integrations_controller.go
```diff
@@ -1933,6 +1950,49 @@ func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zero
...
+	}
+	if len(encTeslaAuth) == 0 {
+		return nil, fmt.Errorf("an error occurred completing tesla device registration")
```
elffjs, line 1970: More specific? No credential found?

internal/controllers/user_integrations_controller.go
```diff
@@ -1933,6 +1950,49 @@ func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zero
...
+	}
+
+	teslaAuth := &services.TeslaAuthCodeResponse{}
+	err = json.Unmarshal([]byte(decrypted), &teslaAuth)
+	if err != nil {
+		udc.log.Err(err).Msgf("failed to unmarshal tesla token found in redis using cacheKey: %s", cacheKey)
+		return nil, errors.Wrap(err, "failed to parse tesla authorization token")
+	}
+
+	if teslaAuth == nil {
```
elffjs, line 1985: Is this possible? It's initialized on 1978.

internal/controllers/user_integrations_controller_test.go
```diff
@@ -892,3 +898,104 @@ func (s *UserIntegrationsControllerTestSuite) TestPairAftermarketNoLegacy() {
 	s.Equal(big.NewInt(4), vID)
 	s.Equal(userSig, u2Sig)
 }
+
+// Tesla Fleet API Tests
+func (s *UserIntegrationsControllerTestSuite) TestPostTesla_FromAuthorizationInRedis() {
```
elffjs, line 903: This test of the "happy case" is good.

I'd like to at least check the case where we can't find anything in Redis. That seems like the most likely thing to happen.

elffjs (changes requested):
Are we going to add a version the job itself in another PR?

internal/controllers/user_integrations_auth_controller.go
```diff
@@ -18,6 +18,8 @@ import (
 	"time"
 )
 
+const TeslaFleetAuthCacheKey = "integration_credentials_%s"
```
KevinJoiner, line 21: Does this need to be exported?
```suggestion
const teslaFleetAuthCacheKey = "integration_credentials_%s"
```

internal/controllers/user_integrations_controller.go
```diff
@@ -1810,6 +1811,24 @@ func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zero
...
 
+	if reqBody.TeslaAPIVersion == 0 {
+		reqBody.TeslaAPIVersion = 1
+	}
+
+	if reqBody.ExternalID == "" {
+		return fiber.NewError(fiber.StatusBadRequest, "Missing externalID parameter")
+	}
+
+	if reqBody.TeslaAPIVersion == 2 { // If version is 2, we are using fleet api which, which has token stored in cache
```
KevinJoiner, line 1822: To avoid magic numbers, we can add constants to the file. You don't have to use these names they are just an example.
``` golang
const (
teslaAPIv1 = 1
teslaAPIv2Fleet = 2
)

File: internal/controllers/user_integrations_controller.go [hunk 10]
```diff
@@ -1933,6 +1953,49 @@ func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zero
 	return c.SendStatus(fiber.StatusNoContent)
 }
 
+func (udc *UserDevicesController) getTeslaAuthFromCache(ctx context.Context, userID string) (*RegisterDeviceIntegrationRequest, error) {
+	user, err := udc.usersClient.GetUser(ctx, &pb.GetUserRequest{Id: userID})
+	if err != nil {
+		return nil, errors.Wrap(err, "could not fetch user information")
```
KevinJoiner, line 1959: We should stop using https://github.com/pkg/errors for new code. This package has been archived for a couple of years since error wrapping was added to the standard library.
 ```suggestion
		return nil, fmt.Errorf("could not fetch user information: %w", err)
```

File: internal/controllers/user_integrations_controller.go [hunk 11]
```diff
@@ -1933,6 +1953,49 @@ func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zero
...
+		udc.log.Err(err).Str("Cache Key", cacheKey).Msg("Error occurred retrieving tesla auth from cache using the key")
+		return nil, errors.Wrap(err, "could not retrieve Tesla credentials")
+	}
+	if len(encTeslaAuth) == 0 {
+		return nil, fmt.Errorf("no credential found")
+	}
+	decrypted, err := udc.cipher.Decrypt(encTeslaAuth)
+	if err != nil {
+		udc.log.Err(err).Str("cacheKey", cacheKey).Msg("failed to decrypt tesla auth from cache using key")
+		return nil, errors.Wrap(err, "failed to decrypt tesla token")
```
KevinJoiner, line 1978: Nit: often it's helpful to either log or return an error but not do both. The reasoning for this is if someone uses this function outside of its current use case they will be forced to log errors even if they do not care about them. Or they will not have meaningful error messages because they were suppressed. I believe we are doing this because we do not want to provide verbose messages in the HTTP response, but it should be up to the function that is handling the response, what gets sent back and what gets logged not the function it is calling.

KevinJoiner (changes requested):
This controller is pretty large. Depending on the level of effort required, it may be worth moving the Tesla stuff to its own package.

internal/services/models.go
```diff
@@ -146,8 +146,9 @@ type UserDeviceAPIIntegrationsMetadata struct {
 	SmartcarUserID          *string                                    `json:"smartcarUserId,omitempty"`
 	Commands                *UserDeviceAPIIntegrationsMetadataCommands `json:"commands,omitempty"`
 	// CANProtocol is the protocol that was detected by edge-network from the autopi.
-	CANProtocol    *string `json:"canProtocol,omitempty"`
-	TeslaVehicleID int     `json:"teslaVehicleId,omitempty"`
+	CANProtocol     *string `json:"canProtocol,omitempty"`
+	TeslaVehicleID  int     `json:"teslaVehicleId,omitempty"`
+	TeslaAPIVersion int     `json:"teslaAPIVersion,omitempty"`
```
elffjs, line 151: This is going to be 0 or... 2? That's a bit strange. Can we leave a comment?

internal/controllers/user_integrations_auth_controller.go
0xdev22: Yes, its being used to form the `cacheKey`

elffjs: But it's only ever being used in this package, yeah?

internal/controllers/user_integrations_controller.go
0xdev22: Normally I would agree, but my intention for this is to decongest the main controller function and not for external use. That being said, there is no reason why a calling function would not want to log errors on the server to help with debugging this further. Also the error isn't being suppressed, if you look at the function its still being returned.

elffjs: I think the ask would be to log as high up as possible, something like line 1825.

internal/services/models.go
elffjs: Oh, and I think I would call it `teslaApiVersion`. Minor thing.

elffjs:
I appreciate the screenshots. In the Redis screenshot, why are the keys all garbled?

elffjs:
> I appreciate the screenshots. In the Redis screenshot, why are the keys all garbled?

Ah, because it's ROT13 locally.

internal/controllers/user_integrations_controller.go
```diff
@@ -1797,6 +1798,7 @@ func (udc *UserDevicesController) registerSmartcarIntegration(c *fiber.Ctx, logg
 }
 
 func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zerolog.Logger, tx *sql.Tx, userDeviceID string, integ *ddgrpc.Integration, ud *models.UserDevice) error {
+	// Flag for which api version should be used
```
elffjs, line 1801: This comment doesn't seem to match what's below it.

internal/controllers/user_integrations_controller.go
```diff
@@ -1810,16 +1812,38 @@ func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zero
...
+	v := &services.TeslaVehicle{}
 	// We'll use this to kick off the job
 	teslaID, err := strconv.Atoi(reqBody.ExternalID)
 	if err != nil {
 		return fiber.NewError(fiber.StatusBadRequest, fmt.Sprintf("Couldn't parse external id %q as an integer.", teslaID))
 	}
 
-	v, err := udc.teslaService.GetVehicle(reqBody.AccessToken, teslaID)
+	region := ""
+	if apiVersion == constants.TeslaAPIV2 { // If version is 2, we are using fleet api which, which has token stored in cache
```
elffjs, line 1832: Slight typo here.

internal/controllers/user_integrations_controller.go
```diff
@@ -1810,16 +1812,38 @@ func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zero
...
+		deviceIntReq, err := udc.getTeslaAuthFromCache(c.Context(), ud.UserID)
+		if err != nil {
+			udc.log.Err(err).Msg("Error occurred retrieving tesla auth from cache")
+			return fiber.NewError(fiber.StatusBadRequest, "Couldn't retrieve stored credentials: "+err.Error())
+		}
+		reqBody.RefreshToken = deviceIntReq.RefreshToken
+		reqBody.AccessToken = deviceIntReq.AccessToken
+		reqBody.ExpiresIn = int(time.Until(deviceIntReq.Expiry).Seconds())
+		region = deviceIntReq.Region
+
```
elffjs, line 1842:
```suggestion
```

internal/services/tesla_fleet_api_service.go
```diff
@@ -91,50 +90,98 @@ func (t *teslaFleetAPIService) CompleteTeslaAuthCodeExchange(ctx context.Context
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("got status code %d waking up vehicle %d", resp.StatusCode, vehicleID)
+	}
+
+	return nil
+}
+
+func teslaHTTPClientHelper(ctx context.Context, logger *zerolog.Logger, httpClient *http.Client, url, token string) (*http.Response, error) {
```
elffjs, line 159: Maybe put this inside `teslaFleetAPIService` so you don't have to pass the logger or HTTP client?

I think the name could be more descriptive. `performTeslaGetRequest`? Something more "active".

internal/services/tesla_fleet_api_service.go
```diff
@@ -91,50 +90,98 @@ func (t *teslaFleetAPIService) CompleteTeslaAuthCodeExchange(ctx context.Context
...
-		return nil, fmt.Errorf("could not fetch vehicles for user: %w", err)
+		return nil, err
 	}
 	req.Header.Set("Authorization", "Bearer "+token)
 
-	resp, err := t.HTTPClient.Do(req)
+	resp, err := httpClient.Do(req)
 	if err != nil {
-		return nil, fmt.Errorf("could not fetch vehicles for user: %w", err)
+		return nil, err
```
elffjs, line 171: I think providing the context in these errors with `fmt.Errorf` was good, we should keep doing it.

internal/services/tesla_fleet_api_service.go
elffjs: Maybe leave a docstring describing what the advantages are here:

1. Adding a timeout to the request
2. Parsing out error responses

internal/services/models.go
elffjs: Bumping this.

internal/services/tesla_task_service.go
```diff
@@ -51,6 +51,7 @@ type TeslaCredentialsV2 struct {
 	AccessToken   string    `json:"accessToken"`
 	Expiry        time.Time `json:"expiry"`
 	RefreshToken  string    `json:"refreshToken"`
+	APIVersion    int       `json:"version"`
```
elffjs, line 54: I would just call them both `version` or both `apiVersion`. These mismatches are a bit odd to me.

internal/controllers/user_integrations_controller_test.go
```diff
@@ -892,3 +902,180 @@ func (s *UserIntegrationsControllerTestSuite) TestPairAftermarketNoLegacy() {
...
+	s.Assert().NoError(err)
+	s.Assert().Equal(1145, oV.ID)
+	s.Assert().Equal(223, oV.VehicleID)
+	s.Assert().Equal(null.StringFrom("1145"), intd.ExternalID)
+	s.Assert().Equal(encAccessToken, intd.AccessToken.String)
+	s.Assert().Equal(encRefreshToken, intd.RefreshToken.String)
+	s.Assert().Equal(2, meta.TeslaAPIVersion)
+}
+
+func (s *UserIntegrationsControllerTestSuite) TestPostTesla_V2_PartialCredentials() {
```
elffjs, line 1010: What makes this "partial"?

elffjs: Ah, there's no refresh token. Very complete. That would certainly be weird.

elffjs (changes requested):
Overall this is looking really good. I think the branching, which was always going to be tough, is handled well. Just minor comments from me.

internal/controllers/user_integrations_controller.go
elffjs: I think this was done, yeah?

internal/services/tesla_fleet_api_service.go
```diff
@@ -91,50 +90,100 @@ func (t *teslaFleetAPIService) CompleteTeslaAuthCodeExchange(ctx context.Context
...
+	if vehicles.Response == nil {
+		return nil, fmt.Errorf("error occurred fetching user vehicles")
+	}
+
+	return vehicles.Response, nil
+}
+
+// GetVehicle
+// @Description Call Tesla Fleet API to get a single vehicle by ID
+func (t *teslaFleetAPIService) GetVehicle(ctx context.Context, token, region string, vehicleID int) (*TeslaVehicle, error) {
```
KevinJoiner, line 119: nit: There is a standard format for function comments so they get rendered correctly as defined here https://tip.golang.org/doc/comment#Funcs
I think this format is only needed if the function is used to generate swagger docs.
```suggestion
// GetVehicle calls Tesla Fleet API to get a single vehicle by ID.
func (t *teslaFleetAPIService) GetVehicle(ctx context.Context, token, region string, vehicleID int) (*TeslaVehicle, error) {
```

internal/controllers/user_integrations_controller.go
```diff
@@ -1810,16 +1811,38 @@ func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zero
 		return fiber.NewError(fiber.StatusBadRequest, "Couldn't parse request body.")
 	}
 
+	// Flag for which api version should be used
+	apiVersion := 1
```
KevinJoiner, line 1815:
```suggestion
	apiVersion := constants.TeslaAPIV1
```

internal/services/tesla_fleet_api_service.go
```diff
@@ -91,50 +90,100 @@ func (t *teslaFleetAPIService) CompleteTeslaAuthCodeExchange(ctx context.Context
...
+	baseURL := fmt.Sprintf(t.Settings.Tesla.FleetAPI, region)
+	url := fmt.Sprintf("%s/api/1/vehicles/%d", baseURL, vehicleID)
```
KevinJoiner, line 121: nit: [path](https://pkg.go.dev/path) is preferred when creating a URLs from varibles.
```suggestion
	url := path.Join(baseURL, "api/1/vehicles", vehicleID)
```

internal/services/tesla_fleet_api_service.go
```diff
@@ -91,50 +90,100 @@ func (t *teslaFleetAPIService) CompleteTeslaAuthCodeExchange(ctx context.Context
...
+
+	resp, err := t.performTeslaGetRequest(ctx, url, token)
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch vehicles for user: %w", err)
+	}
+	defer resp.Body.Close()
+
+	vehicle := new(struct {
+		Response TeslaVehicle `json:"response"`
+	})
```
KevinJoiner, line 131: We can move this struct definition outside the function since we don't need to define the struct every time it is run.

KevinJoiner (changes requested):
This looks good only some small comments from me.

internal/services/tesla_fleet_api_service.go
elffjs: Any advantage to

https://pkg.go.dev/net/url#JoinPath

?

elffjs: > nit: [path](https://pkg.go.dev/path) is preferred when creating a URLs from varibles.

I guess you would have to convert `vehicleID` to a string, first.

0xdev22: Since this is a nit, I'd rather stick to `fmt.Sprintf`, matter of preference.

internal/services/tesla_task_service.go
```diff
@@ -51,6 +51,7 @@ type TeslaCredentialsV2 struct {
 	AccessToken   string    `json:"accessToken"`
 	Expiry        time.Time `json:"expiry"`
 	RefreshToken  string    `json:"refreshToken"`
+	APIVersion    int       `json:"apiVersion"`
```
elffjs, line 54: It's confusing to me that it's `version` in one place, `apiVersion` in another, `teslaApiVersion` in a third. I'm going to cope and say that this is temporary and that doesn't matter much. Let's see.

elffjs (approved):

elffjs (approved):

internal/services/tesla_task_service.go
elffjs: @KevinJoiner Any opinion?

KevinJoiner: I agree. I had the same confusion. I wasn't sure if `RegisterDeviceIntegrationRequest` was only used for Tesla or not. If we also use this request for synthetic devices, then simply using the version works.

internal/services/tesla_fleet_api_service.go
KevinJoiner: unresolving this because it wasn't addressed.

KevinJoiner (changes requested):

internal/services/tesla_fleet_api_service.go
0xdev22: https://github.com/DIMO-Network/devices-api/blob/88a934688599f3bdf8d3be47034b1de3fd1d67b3/internal/services/tesla_fleet_api_service.go#L130

KevinJoiner (approved):

PR #271, opened 2024-02-27
SI-2471-burn-vehicle

Allyson-English:
- add GET and POST endpoints to burn a minted vehicle
   -  GET /v1/vehicle/:tokenID/commands/burn
   -  POST /v1/vehicle/:tokenID/commands/burn
- handler for VehicleBurnedEvent in contract event consumer
- update `udOwner.Delete("/", userDeviceController.DeleteUserDevice)`; check if device has been minted; if yes, return an error (StatusFailedDependency, 424) and instruct the user to burn the device
- tests

NOTES:

- do we want to keep `/commands` in the burn endpoint path? This is consistent with what we have elsewhere but it could also make sense to make the path `/v1/vehicle/:tokenID/burn`
- is StatusFailedDependency (424) the best error code to return from the DeleteUserDevice endpoint? I was aiming to make it distinct from other errors returned by that function to make it easier for the frontend to handle


elffjs:
Do we have a handler for `VehicleNodeBurned` already? Maybe that should be the first PR, now that I think about it.

internal/services/registry/storage.go
```diff
@@ -138,6 +140,42 @@ func (p *proc) Handle(ctx context.Context, data *ceData) error {
...
+		for _, l1 := range data.Transaction.Logs {
+			if l1.Topics[0] == vehicleBurnedEvent.ID {
+				out := new(contracts.RegistryVehicleNodeBurned)
+				err := p.parseLog(out, vehicleBurnedEvent, l1)
+				if err != nil {
+					return err
+				}
+
+				vnft := mtr.R.BurnRequestVehicleNFT
+				if _, err := vnft.Delete(ctx, p.DB().Writer); err != nil {
```
elffjs, line 153: Should we not also delete the `user_device`?

internal/services/registry/storage.go
```diff
@@ -138,6 +140,42 @@ func (p *proc) Handle(ctx context.Context, data *ceData) error {
 				}
 			}
 		}
+	case mtr.R.BurnRequestVehicleNFT != nil:
```
elffjs, line 143: I would imitate the claim handler and have this basically be a no-op.

Put the burn "handler" in the contract event processor so that you can handle the raw event if someone burns directly on-chain.

internal/controllers/user_devices_controller.go
```diff
@@ -1856,6 +1856,173 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
+		Status: "Unsubmitted",
+	}
+	err = mtr.Insert(c.Context(), udc.DBS().Writer, boil.Infer())
+	if err != nil {
+		return err
+	}
+
+	userDevice.R.VehicleNFT.BurnRequestID = null.StringFrom(requestID)
+
+	if _, err := userDevice.R.VehicleNFT.Update(c.Context(), udc.DBS().Writer, boil.Infer()); err != nil {
```
elffjs, line 2018: Probably want to be careful with a transaction here. You do not want someone to double-submit this. See the claim handler for an example.

internal/services/registry/client.go
```diff
@@ -311,6 +332,21 @@ func (c *Client) MintVehicleAndSDign(requestID string, data contracts.MintVehicl
 	return c.sendRequest(requestID, callData)
 }
 
+// bintVehicleSign(uint256 manufacturerNode, address owner,	string[] calldata attributes, string[] calldata infos, bytes calldata signature)
```
elffjs, line 335: bint?

internal/controllers/nft_controller.go
```diff
@@ -837,3 +840,200 @@ func (nc *NFTController) GetVinCredential(c *fiber.Ctx) error {
...
+	tis := c.Params("tokenID")
+	ti, ok := new(big.Int).SetString(tis, 10)
+	if !ok {
+		return fiber.NewError(fiber.StatusBadRequest, fmt.Sprintf("Couldn't parse token id %q.", tis))
+	}
+	tid := types.NewNullDecimal(new(decimal.Big).SetBigMantScale(ti, 0))
+
+	client := registry.Client{
+		Producer:     udc.producer,
+		RequestTopic: "topic.transaction.request.send",
```
KevinJoiner, line 861: Should `topic.transaction.request.send` be a setting similar to some of our other topics.

internal/controllers/nft_controller.go
```diff
@@ -837,3 +840,200 @@ func (nc *NFTController) GetVinCredential(c *fiber.Ctx) error {
...
+	vehicleNFT, err := models.VehicleNFTS(
+		models.VehicleNFTWhere.TokenID.EQ(tid),
+		qm.Load(qm.Rels(models.VehicleNFTRels.UserDevice)),
+		qm.Load(qm.Rels(models.VehicleNFTRels.MintRequest)),
+		qm.Load(qm.Rels(models.VehicleNFTRels.VehicleTokenAftermarketDevice)),
+		qm.Load(qm.Rels(models.VehicleNFTRels.VehicleTokenSyntheticDevice)),
+		qm.Load(qm.Rels(models.VehicleNFTRels.UserDevice, models.UserDeviceRels.UserDeviceAPIIntegrations)),
+	).One(c.Context(), tx)
+	if err != nil {
+		return fiber.NewError(fiber.StatusBadRequest, "failed to find device by token id: %s", tis)
```
KevinJoiner, line 885: If the NFT does not exist, then we should return a 404 since the object is not found. Otherwise, we can do a 500 and log the error. I think you can check if the errors.Is(sql.NoRow) to determine if it should be 404 or 500

internal/controllers/nft_controller.go
```diff
@@ -837,3 +840,200 @@ func (nc *NFTController) GetVinCredential(c *fiber.Ctx) error {
...
+	vehicleNFT.BurnRequestID = null.StringFrom(requestID)
+	if _, err := vehicleNFT.Update(c.Context(), tx, boil.Infer()); err != nil {
+		return fiber.NewError(fiber.StatusBadRequest, "failed to associate burn request with vehicle")
+	}
+
+	if err := tx.Commit(); err != nil {
+		return err
+	}
+
+	udc.log.Info().Msgf("submitted metatransaction request %s", requestID)
```
KevinJoiner, line 1003: nit:
Maybe drop this down to debug since we do not need to know every time this function is run.

KevinJoiner (changes requested):
I didn't finish looking at this, but I took a first glance. I think 424 makes sense as a return code.

internal/controllers/nft_controller.go
Allyson-English: hesitant to do that bc that topic is used across rewards api, meta tx processor, etc... but it probably should be coming from settings at the very least rather than hardcoded !

Allyson-English: follow up q: if we made this change for this endpoint, we'd probably want to carry the change through to where this topic is published to elsewhere in the repo... should that be a separate PR making that change everywhere at once so that we don't clutter up this pr?

KevinJoiner (changes requested):

internal/services/registry/client.go
```diff
@@ -311,6 +332,21 @@ func (c *Client) MintVehicleAndSDign(requestID string, data contracts.MintVehicl
 	return c.sendRequest(requestID, callData)
 }
 
+// BurnVehicleSign(uint256 manufacturerNode, address owner,	string[] calldata attributes, string[] calldata infos, bytes calldata signature)
```
KevinJoiner, line 335: update

KevinJoiner (changes requested):

KevinJoiner (approved):

PR #272, opened 2024-02-29
Delete teslaAuth tokens from cache after persisting to kafka and db

0xdev22:
# Proposed Changes


internal/controllers/user_integrations_controller.go
```diff
@@ -1829,8 +1829,18 @@ func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zero
 	}
 
 	region := ""
+	teslaV2CacheKey := ""
 	if apiVersion == constants.TeslaAPIV2 { // If version is 2, we are using fleet api which has token stored in cache
-		deviceIntReq, err := udc.getTeslaAuthFromCache(c.Context(), ud.UserID)
+		user, err := udc.usersClient.GetUser(c.Context(), &pb.GetUserRequest{Id: ud.UserID})
```
elffjs, line 1834: Okay, we're moving this up in order to provide nicer messages.

elffjs (approved):

internal/controllers/user_integrations_controller.go
0xdev22: Not really, Its because I need the cache key for deletion and its formed using the user eth address. But that is also a good point, I will use it.

KevinJoiner: 2 🐦  1 🪨

KevinJoiner (approved):

PR #277, opened 2024-03-12
Create grpc endpoint to clear stuck meta transaction request

0xdev22:
# Proposed Changes


internal/rpc/user_devices.go
```diff
@@ -635,3 +637,21 @@ func (s *userDeviceRPCServer) UpdateUserDeviceMetadata(ctx context.Context, req
 
 	return &emptypb.Empty{}, nil
 }
+
+func (s *userDeviceRPCServer) ClearMetaTransactionRequests(ctx context.Context, _ *emptypb.Empty) (*pb.ClearMetaTransactionRequestsResponse, error) {
+	currTime := time.Now()
+	fifteenminsAgo := currTime.Add(-time.Minute * 15)
+
+	m, err := models.MetaTransactionRequests(
+		models.MetaTransactionRequestWhere.CreatedAt.LT(fifteenminsAgo),
```
KevinJoiner, line 646: Since this is not an automated step, I don't think we need to guarantee that the request has been stuck for 15 minutes.

0xdev22: [https://linear.app/dimo/issue/SI-2496/clear-stuck-transactions-automatically-after-15-minutes
](https://linear.app/dimo/issue/SI-2496/clear-stuck-transactions-automatically-after-15-minutes
)
The ticket Seems to suggest that, no other parameter were given on conditions to clear the transaction other than the length of time.

internal/rpc/user_devices_test.go
```diff
@@ -250,3 +253,113 @@ func TestGetUserDevice_NoSyntheticDeviceFields_WhenNoTokenID(t *testing.T) {
...
+	userDeviceSvc := services.NewUserDeviceService(nil, logger, pdb.DBS, nil, nil)
+	udService := NewUserDeviceRPCService(pdb.DBS, nil, nil, nil, nil, nil, nil, userDeviceSvc)
+
+	resp, err := udService.ClearMetaTransactionRequests(ctx, nil)
+	assert.NoError(err)
+
+	assert.Equal(mtID, resp.Id)
+
+	_, err = models.MetaTransactionRequests(models.MetaTransactionRequestWhere.ID.EQ(mtID)).One(ctx, pdb.DBS().Reader)
+	assert.ErrorIs(err, sql.ErrNoRows)
```
KevinJoiner, line 290: Does this pass? we are not wrapping the error when the err == sql.ErrNoRows

KevinJoiner (changes requested):

internal/rpc/user_devices_test.go
0xdev22: Yes it does, sqlboiler throws a `sql.ErrNoRows`when there are no records matching the condition. What i'm doing here is retrieving the record directly using an sqlboiler model, its not a custom function so no wrapping.

internal/rpc/user_devices.go
0xdev22: I think it was quite clear though, that the time it has been on queue is the most important thing. But anyways, I believe this has been clarified.

internal/rpc/user_devices_test.go
KevinJoiner: Ahh yes, my confusion I misread the function call above and thought it was `ClearMetaTransactionRequests`

KevinJoiner (approved):
Your changes look good, but it looks like CI is still failing.

PR #285, opened 2024-03-22
Add vehicle to tesla telemetry server with tests

0xdev22:
# Proposed Changes

The PR adds the following:
- Fleet API function to help register a vehicle with the tesla telemetry [endpoint](https://developer.tesla.com/docs/fleet-api#fleet_status) 
- Add vehicle to telemetry server once Tesla registration is done
- Add complete set of fields to be reported in telemetry
- Test to verify how tesla telemetry API is being called
- Add non-secret config data to environment variables

### Impacted Routes
`/user/devices/:userDeviceID/integrations/:integrationID [post]
`


Allyson-English (approved):

internal/services/tesla_fleet_api_service.go
```diff
@@ -155,6 +176,66 @@ func (t *teslaFleetAPIService) WakeUpVehicle(ctx context.Context, token, region
 	return nil
 }
 
+func (t *teslaFleetAPIService) RegisterToTelemetryServer(ctx context.Context, token, region, vin string) error {
+	baseURL := fmt.Sprintf(t.Settings.TeslaFleetURL, region)
+	u, err := url.Parse(fmt.Sprintf("%s/api/1/vehicles/fleet_telemetry_config", baseURL))
+	if err != nil {
+		return err
+	}
```
KevinJoiner, line 184: Since we do not have control over the baseURL setting we should be able to handle if there is a trailing slash or not this can be done by using URL.JoinPath(). I wrote an example of how a trailing slash breaks this code
https://go.dev/play/p/ChKqJMYcrgt

internal/controllers/user_integrations_controller.go
```diff
@@ -1948,6 +1949,14 @@ func (udc *UserDevicesController) registerDeviceTesla(c *fiber.Ctx, logger *zero
 		return err
 	}
 
+	if apiVersion == constants.TeslaAPIV2 {
+		err = udc.registerTeslaForTelemetryServer(c.Context(), region, reqBody.AccessToken, v)
+		if err != nil {
+			// We should consider pushing this to a Queue for retry
+			logger.Err(err).Str("vin", v.VIN).Int("id", v.ID).Msg("Could not register Tesla for Telemetry")
```
KevinJoiner, line 1956: I don't think we want to log a mapping between people's VIN and their vehicle ID since these logs get stored in elastic with the vehicle data. where we specifically strip out the VIN.

KevinJoiner (changes requested):

charts/devices-api/values.yaml
```diff
@@ -94,6 +94,9 @@ env:
...
...udCBSb290MCAXDTIyMDQyMzExMTEw\nM1oYDzIwNzIwNDEwMTExMTAzWjAyMTAwLgYDVQQDEydESU1PIENBIERldmVsb3Bt\nZW50IFNlcnZlciBJbnRlcm1lZGlhdGUwWTATBgcqhkjOPQIBBggqhkjOPQMBBwNC\nAAQMyh6plUM3p4KWWfK0CqWXr1B9NWk53+c9ps8OpgZZIyXjxiw1EHxrpcqU7C9e\nhw+6JfmvTqqi3F4ES8K+Tt/mo2YwZDAOBgNVHQ8BAf8EBAMCAQYwEgYDVR0TAQH/\nBAgwBgEB/wIBADAdBgNVHQ4EFgQU+7zrfioO4bjNpD9KiG8fbTcIq8kwHwYDVR0j\nBBgwFoAUeMfSSqt+S65xQF82yRnjr+J5XC8wCgYIKoZIzj0EAwIDSAAwRQIhAK3s\nWtlk+d0fnkii091dTZGt+dtzEbM4HuizaG6mO5zPAiApi03qU/hdsAxXwlbhufH/\n5HuUiCLgBK8vPvL2YdMaKQ==\n-----END CERTIFICATE-----\n-----BEGIN CERTIFICATE-----\nMIIBrTCCAVKgAwIBAgIQEgthFz9Ww3+VaErBc3nDFjAKBggqhkjOPQQDAjAjMSEw\nHwYDVQQDExhESU1PIENBIERldmVsb3BtZW50IFJvb3QwIBcNMjIwNDIzMTExMTAz\nWhgPMjEyMjAzMzAxMTExMDNaMCMxITAfBgNVBAMTGERJTU8gQ0EgRGV2ZWxvcG1l\nbnQgUm9vdDBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABBuxEb6jTyfrUwI8RiBV\nKCQWqTAeLdHPj60Qk7HeMeaEcGjzF799xgpl6/8iNKaHN/w+705cdxp5pRswbUtu\nizWjZjBkMA4GA1UdDwEB/wQEAwIBBjASBgNVHRMBAf8ECDAGAQH/AgEBMB0GA1Ud\nDgQWBBR4x9JKq35LrnFAXzbJGeOv4nlcLzAfBgNVHSMEGDAWgBR4x9JKq35LrnFA\nXzbJGeOv4nlcLzAKBggqhkjOPQQDAgNJADBGAiEAlslTE9mX+VjPSYLKEsy48Rzh\nOUCdaWovmF+28PyAi4wCIQDXRKpYK+VMFyUR1GJVoV3gWezQcJmFswuWq+7M+XPb\nGQ==\n-----END CERTIFICATE-----
```
JamesReate, line 99: just want to check that this should not be a secret? looks like something we may want to be a secret

internal/services/tesla_fleet_api_service.go
0xdev22: Not sure I understand what you mean by not having control over the baseURL setting

charts/devices-api/values.yaml
0xdev22: Its fine, its supposed to be a public CA certificate https://developer.tesla.com/docs/fleet-api#fleet_status:~:text=public%20certificate%20authority%20cert%20for%20fleet%2Dtelemetry%20server

internal/services/tesla_fleet_api_service.go
```diff
@@ -155,6 +176,66 @@ func (t *teslaFleetAPIService) WakeUpVehicle(ctx context.Context, token, region
...
+			PublicCACertificate: t.Settings.TeslaTelemetryCACertificate,
+			Expiration:          exp,
+			Port:                t.Settings.TeslaTelemetryPort,
+			Fields:              make(map[string]Interval),
+			AlertTypes:          []string{"service"},
+		},
+	}
+	for _, v := range fields {
+		r.Config.Fields[v] = Interval{
+			IntervalSeconds: 1800,
```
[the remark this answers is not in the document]
0xdev22: Its a good observation. We can use thesame server though and just change the interval for the ones that are more important, of course we will have to determine which ones are more important.

internal/services/tesla_fleet_api_service.go
KevinJoiner: The baseURL comes from a setting file that is read in at runtime, so this function has no control over whether or not there is a trailing slash. Using Sprintf like you have now is what caused this bug in identity API https://linear.app/dimo/issue/SI-2524/identity-api-image-creates-a-malformed-url

0xdev22: so the base url looks like this `https://fleet-api.prd.%s.vn.cloud.tesla.com` where `%s`  a **region** argument passed from the client. So we still need to use the `fmt.Sprintf`. However, it might also be a good idea to use url.Join

KevinJoiner (approved):

internal/services/tesla_fleet_api_service.go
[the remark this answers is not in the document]
zer0stars: @JamesReate we can specify a different config on a per-vehicle basis.

internal/controllers/user_integrations_controller.go
```diff
@@ -72,7 +73,7 @@ func (udc *UserDevicesController) deleteDeviceIntegration(ctx context.Context, u
 	if err != nil {
 		return err
 	}
-	defer tx.Rollback() //nolint
+	defer tx.Rollback() // nolint
```
elffjs, line 76: Having no space here is correct. It's a comment for machines.

elffjs:
Trying to tell the difference between this and https://github.com/DIMO-Network/devices-api/pull/315

PR #290, opened 2024-03-26
[SM-703] add depencies delete for stuck meta transactions

Asfiroth:
# Proposed Changes
- Add delete for vehicle nft relation
- Add delete for synthetic device relation

### DIFF
--- a/internal/rpc/user_devices.go
+++ b/internal/rpc/user_devices.go
@@ -642,10 +642,11 @@ func (s *userDeviceRPCServer) ClearMetaTransactionRequests(ctx context.Context, 
 	currTime := time.Now()
 	fifteenminsAgo := currTime.Add(-time.Minute * 15)
 
-	m, err := models.MetaTransactionRequests(
+	metaTransaction, err := models.MetaTransactionRequests(
 		models.MetaTransactionRequestWhere.CreatedAt.LT(fifteenminsAgo),
 		qm.OrderBy(models.MetaTransactionRequestColumns.CreatedAt+" ASC"),
 	).One(ctx, s.dbs().Reader)
+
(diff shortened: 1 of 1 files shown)


internal/config/settings.go
```diff
@@ -78,6 +78,8 @@ type Settings struct {
 
 	DIMOContractAPIURL string `yaml:"DIMO_CONTRACT_APIURL"`
 
+	MetaTransactionProcessorGRPCAddr string `yaml:"META_TRANSACTION_PROCESSOR_GRPC_ADDR"`
```
JamesReate, line 81: do we need something in values.yaml?

internal/rpc/user_devices.go
```diff
@@ -639,24 +641,119 @@ func (s *userDeviceRPCServer) UpdateUserDeviceMetadata(ctx context.Context, req
...
+			_, err = vehicleNft.Update(ctx, s.dbs().Writer, boil.Infer())
+			if err != nil {
+				return nil, fmt.Errorf("failed to update vehicleNft %s: %w", vehicleNft.TokenID, err)
+			}
+		}
+
+		if metaTransaction.R.BurnRequestVehicleNFT != nil {
+			vehicleNft := metaTransaction.R.BurnRequestVehicleNFT
+
+			vehicleNft.BurnRequestID = null.StringFrom("")
```
JamesReate, line 695: is there an option for null string, or null.New, maybe even null.StringFromPtr(nil)

internal/config/settings.go
Asfiroth: yup, I forgot to add it

JamesReate (approved):

PR #301, opened 2024-04-02
SI-2525-disconnect-synthetic-vehicles-programatically

Allyson-English:
Add `StopUserDeviceIntegration` gRPC endpoint so we can programmatically disconnect all the Kia's that have been added. This will put the Smartcar integrations into an `AuthenticationFailed` state so, in the event we cast too broad of a net, the valid user will simply have to re-authenticate. 

**The is editing an earlier draft of this, which is already in main bc I was testing it, that deleted the integration. This change will just require re-auth.**


internal/rpc/user_devices.go
```diff
@@ -762,86 +760,63 @@ func (s *userDeviceRPCServer) ClearMetaTransactionRequests(ctx context.Context,
...
+		log.Info().Str("vendor", integ.Vendor).Msg("stop user integration poll not implemented for vendor")
+		return nil, fmt.Errorf("stop user integration poll not implemented for vendor %s", integ.Vendor)
+	}
 
-		_, err = apiInt.Delete(ctx, s.dbs().Reader)
-		if err != nil {
-			return nil, fmt.Errorf("failed to delete integration: %w", err)
-		}
+	apiInt.Status = models.UserDeviceAPIIntegrationStatusAuthenticationFailure
+	apiInt.TaskID = null.String{}
```
0xdev22, line 813: I don't think we should remove the TaskID, it can still be used for reference. I think changing the status is good enough.

0xdev22 (changes requested):

0xdev22 (approved):

PR #303, opened 2024-04-09
SI-2472-merge-vnft-into-ud-tbl

Allyson-English:
# Proposed Changes
Merge `user_devices` and `vehicle_nfts` tables
Vehicles can be unminted but burned vehicles should be deleted everywhere (this should be done after this migration is applied so that we are burning orphaned vehicles not just removing them from our db)


KevinJoiner (commented):
Didn't finish

migrations/20240405110006_vnft_into_ud_tbl.sql
```diff
@@ -0,0 +1,128 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query';
+
+SET search_path = devices_api, public;
+CREATE TABLE vehicle_nfts_backup AS SELECT *  FROM vehicle_nfts;
```
elffjs, line 6: What does this do?

migrations/20240405110006_vnft_into_ud_tbl.sql
```diff
@@ -0,0 +1,128 @@
...
+        burn_request_id = vnft.burn_request_id,
+        mint_request_id = vnft.mint_request_id,
+        token_id = vnft.token_id,
+        claim_id = vnft.claim_id,
+        owner_address = vnft.owner_address
+FROM vehicle_nfts vnft 
+WHERE vnft.user_device_id = user_devices.id;
+
+ALTER TABLE aftermarket_devices DROP CONSTRAINT autopi_units_vehicle_token_id_fkey;
+ALTER TABLE aftermarket_devices ADD CONSTRAINT autopi_units_vehicle_token_id_fkey FOREIGN KEY (vehicle_token_id) REFERENCES user_devices(token_id) ON DELETE CASCADE;
```
elffjs, line 34: Maybe on set null?

migrations/20240405110006_vnft_into_ud_tbl.sql
```diff
@@ -0,0 +1,128 @@
...
+
+ALTER TABLE synthetic_devices DROP CONSTRAINT fkey_vehicle_token_id;
+ALTER TABLE synthetic_devices ADD CONSTRAINT fkey_vehicle_token_id FOREIGN KEY (vehicle_token_id) REFERENCES user_devices(token_id);
+
+DROP TABLE vehicle_nfts;
+DROP TABLE vehicle_nfts_backup;
+
+CREATE MATERIALIZED VIEW vehicle_nfts
```
elffjs, line 42: Would a (non-materialized) view be much worse for our use case?

migrations/20240405110006_vnft_into_ud_tbl.sql
```diff
@@ -0,0 +1,128 @@
...
+
+ALTER TABLE user_devices
+    ADD COLUMN mint_request_id char(27) CONSTRAINT user_devices_mint_request_id_fkey UNIQUE REFERENCES meta_transaction_requests(id);
+
+ALTER TABLE user_devices
+    ADD COLUMN burn_request_id char(27) CONSTRAINT user_devices_burn_request_id_fkey UNIQUE REFERENCES meta_transaction_requests(id);
+
+ALTER TABLE user_devices
+    ADD COLUMN token_id numeric(78, 0) CONSTRAINT user_devices_token_id_key UNIQUE;
+
```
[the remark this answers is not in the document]
elffjs: I _believe_ unique columns get an index automatically.

migrations/20240405110006_vnft_into_ud_tbl.sql
```diff
@@ -0,0 +1,128 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query';
+
```
elffjs, line 4:
```suggestion
```

internal/controllers/user_devices_controller.go
```diff
@@ -356,17 +354,16 @@ func (udc *UserDevicesController) GetUserDevices(c *fiber.Ctx) error {
...
+			qm.Or2(models.UserDeviceWhere.OwnerAddress.EQ(null.BytesFrom(common.HexToAddress(*user.EthereumAddress).Bytes()))),
 		}
 	}
 
 	query = append(query,
 		qm.Load(models.UserDeviceRels.UserDeviceAPIIntegrations),
-		qm.Load(qm.Rels(models.UserDeviceRels.VehicleNFT, models.VehicleNFTRels.MintRequest)),
-		qm.Load(qm.Rels(models.UserDeviceRels.VehicleNFT, models.VehicleNFTRels.VehicleTokenSyntheticDevice, models.SyntheticDeviceRels.MintRequest)),
-		qm.Load(qm.Rels(models.UserDeviceRels.VehicleNFT, models.VehicleNFTRels.Claim)),
+		qm.Load(qm.Rels(models.UserDeviceRels.MintRequest)),
```
elffjs, line 364: Can get rid of `qm.Rels` here and two lines down.

internal/controllers/user_integrations_controller.go
```diff
@@ -173,7 +173,7 @@ func (udc *UserDevicesController) DeleteUserDeviceIntegration(c *fiber.Ctx) erro
 	device, err := models.UserDevices(
 		models.UserDeviceWhere.ID.EQ(userDeviceID),
 		qm.Load(models.UserDeviceRels.UserDeviceAPIIntegrations, models.UserDeviceAPIIntegrationWhere.IntegrationID.EQ(integrationID)),
-		qm.Load(qm.Rels(models.UserDeviceRels.VehicleNFT, models.VehicleNFTRels.VehicleTokenSyntheticDevice)),
+		qm.Load(qm.Rels(models.UserDeviceRels.VehicleTokenSyntheticDevice)),
```
elffjs, line 176: Can delete `Rels`.

internal/rpc/user_devices.go
```diff
@@ -80,19 +80,16 @@ func (s *userDeviceRPCServer) GetUserDevice(ctx context.Context, req *pb.GetUser
 	dbDevice, err := models.UserDevices(
 		models.UserDeviceWhere.ID.EQ(req.Id),
 		qm.Load(
-			qm.Rels(models.UserDeviceRels.VehicleNFT,
-				models.VehicleNFTRels.VehicleTokenAftermarketDevice),
+			qm.Rels(models.UserDeviceRels.VehicleTokenAftermarketDevice),
```
elffjs, line 83: Drop rels a few places here too.

elffjs:
I think we can get rid of a lot of `qm.Rels` and I wonder about the necessity of a materialized view. Line by line it seems to work!

migrations/20240405110006_vnft_into_ud_tbl.sql
Allyson-English: I was thinking it would be faster to have a materialized view but I think you're right-- since we would have to refresh it all the time anyways, a regular view probably makes more sense

elffjs: My unscientific argument is that it's kinda just `user_devices` with some columns removed.

migrations/20240501024340_vnft_into_ud_tbl.sql
```diff
@@ -0,0 +1,109 @@
+-- +goose Up
+-- +goose StatementBegin
+SET search_path = devices_api, public;
+
+ALTER TABLE user_devices
+    ADD COLUMN mint_request_id char(27) CONSTRAINT user_devices_mint_request_id_fkey UNIQUE REFERENCES meta_transaction_requests(id);
```
elffjs, line 6: I didn't know you could combine and the foreign key. Could we separate these out? Give them both names (Postgres will give them names anyway, I just like having them written out).

migrations/20240501024340_vnft_into_ud_tbl.sql
```diff
@@ -0,0 +1,109 @@
...
+
+UPDATE user_devices 
+    SET 
+        burn_request_id = vnft.burn_request_id,
+        mint_request_id = vnft.mint_request_id,
+        token_id = vnft.token_id,
+        claim_id = vnft.claim_id,
+        owner_address = vnft.owner_address
+FROM vehicle_nfts vnft 
+WHERE vnft.user_device_id = user_devices.id;
```
elffjs, line 28: Note that we still have some rows in `vehicle_nfts` with null `user_device_id`. We should look into those.

migrations/20240501024340_vnft_into_ud_tbl.sql
```diff
@@ -0,0 +1,109 @@
...
+
+ALTER TABLE aftermarket_devices DROP CONSTRAINT autopi_units_vehicle_token_id_fkey;
+ALTER TABLE aftermarket_devices ADD CONSTRAINT autopi_units_vehicle_token_id_fkey FOREIGN KEY (vehicle_token_id) REFERENCES user_devices(token_id) ON DELETE SET NULL;
```
elffjs, line 31: Could we take the opportunity to correct the name of this constraint? `aftermarket_devices_vehicle_token_id_fkey` or something.

migrations/20240501024340_vnft_into_ud_tbl.sql
```diff
@@ -0,0 +1,109 @@
...
+        ud.id as user_device_id,
+        ud.mint_request_id,
+        ud.token_id,
+        ud.claim_id,
+        ud.owner_address,
+        ud.vin_identifier as vin
+FROM user_devices ud;
+
+ALTER TABLE aftermarket_devices DROP CONSTRAINT autopi_units_vehicle_token_id_fkey;
+ALTER TABLE aftermarket_devices ADD CONSTRAINT autopi_units_vehicle_token_id_fkey FOREIGN KEY (vehicle_token_id) REFERENCES vehicle_nfts(token_id) ON DELETE CASCADE;
```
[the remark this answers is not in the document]
Allyson-English: what do you think about removing ON DELETE all together? it will throw an error if we haven't unpaired first (although that shouldnt happen)

migrations/20240501024340_vnft_into_ud_tbl.sql
```diff
@@ -0,0 +1,109 @@
+-- +goose Up
```
elffjs, line 1: @JamesReate How scary does this migration look to you? Do you think Admin will throw a fit?

elffjs:
Going to try to merge in main. Unfortunately a lot of migrations have been added in the meantime.

elffjs:
Okay, I think that worked. Tests pass and all that.

internal/controllers/nft_controller.go
```diff
@@ -215,12 +200,7 @@ func (nc *NFTController) GetNFTImage(c *fiber.Ctx) error {
 		return opaqueInternalError
 	}
 
-	if nft.R.UserDevice == nil {
-		helpers.SkipErrorLog(c)
-		return fiber.NewError(fiber.StatusNotFound, "NFT not found.")
-	}
-
-	imageName = nft.MintRequestID
+	imageName := nft.MintRequestID.String
```
elffjs, line 203: The old code assumes that `mint_request_id` is always populated, since that's the primary key of the table; this is no longer a safe assumption. And in fact right now we're not able to provide an image for vehicles whose mint calls were directly to the chain.

I think if `mint_request_id` is null then we should error here.

elffjs: Fixed it, I think.

internal/controllers/nft_controller.go
```diff
@@ -634,11 +613,12 @@ func (udc *UserDevicesController) GetBurnDevice(c *fiber.Ctx) error {
...
 	defer tx.Rollback() //nolint
 
-	vehicleNFT, err := models.VehicleNFTS(
-		models.VehicleNFTWhere.TokenID.EQ(tid),
-		qm.Load(models.VehicleNFTRels.UserDevice),
-		qm.Load(models.VehicleNFTRels.VehicleTokenAftermarketDevice),
-		qm.Load(models.VehicleNFTRels.VehicleTokenSyntheticDevice),
+	userDevice, err := models.UserDevices(
+		models.UserDeviceWhere.TokenID.EQ(tid),
+		qm.Load(models.UserDeviceRels.BurnRequest),
```
elffjs, line 618: This was actually a bug of mine on main, whoops.

internal/controllers/user_devices_controller.go
```diff
@@ -231,28 +231,26 @@ func (udc *UserDevicesController) dbDevicesToDisplay(ctx context.Context, device
 		var nft *VehicleNFTData
 		var credential *VINCredentialData
 		pu := []PrivilegeUser{}
-
-		if vnft := d.R.VehicleNFT; vnft != nil {
-			nftStatus := vnft.R.MintRequest
+		if !d.TokenID.IsZero() {
```
elffjs, line 234: Unfortunately I think we have to be more careful here. There can still be interesting information in this section even if minting isn't finished (and so a token id is present).

internal/services/fingerprint/consumer.go
```diff
@@ -107,67 +107,64 @@ func (c *Consumer) HandleDeviceFingerprint(ctx context.Context, event *Event) er
 
 	ad, err := models.AftermarketDevices(
 		models.AftermarketDeviceWhere.EthereumAddress.EQ(addr.Bytes()),
-		qm.Load(qm.Rels(models.AftermarketDeviceRels.VehicleToken, models.VehicleNFTRels.Claim)),
+		qm.Load(models.AftermarketDeviceRels.VehicleToken),
```
elffjs, line 110: This doesn't look safe.

internal/controllers/user_devices_controller.go
```diff
@@ -349,17 +352,16 @@ func (udc *UserDevicesController) GetUserDevices(c *fiber.Ctx) error {
 		}
 	} else {
 		query = []qm.QueryMod{
-			qm.LeftOuterJoin("devices_api." + models.TableNames.VehicleNFTS + " ON " + models.VehicleNFTTableColumns.UserDeviceID + " = " + models.UserDeviceTableColumns.ID),
```
elffjs, line 352: Nice to clean this up.

internal/controllers/user_devices_controller.go
```diff
@@ -1192,7 +1193,7 @@ func (udc *UserDevicesController) DeleteUserDevice(c *fiber.Ctx) error {
 	}
 
 	// if vehicle minted, user must delete by burning
-	if !userDevice.TokenID.IsZero() {
+	if !userDevice.TokenID.IsZero() || userDevice.R.MintRequest != nil && userDevice.R.MintRequest.Status != models.MetaTransactionRequestStatusFailed {
```
elffjs, line 1196: Don't let them delete if minting is in progress.

internal/controllers/user_devices_controller.go
```diff
@@ -1333,12 +1334,11 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
-		switch userDevice.R.MintRequest.Status {
-		case "Confirmed":
-			return fiber.NewError(fiber.StatusConflict, "Vehicle already minted.")
-		default:
-			return fiber.NewError(fiber.StatusConflict, "Minting in process.")
-		}
+		return fiber.NewError(fiber.StatusConflict, "Vehicle already minted.")
+	}
+
+	if mr := userDevice.R.MintRequest; mr != nil && mr.Status != models.MetaTransactionRequestStatusFailed {
```
elffjs, line 1340: Don't let them mint if there's already one in progress.

internal/controllers/user_devices_controller.go
```diff
@@ -1587,6 +1587,11 @@ func (udc *UserDevicesController) UpdateNFTImage(c *fiber.Ctx) error {
 		return fiber.NewError(fiber.StatusBadRequest, "Vehicle not minted.")
 	}
 
+	// This would only happen if it's a "permissionless mint".
+	if !userDevice.MintRequestID.Valid {
```
elffjs, line 1591: I named the images in S3 using the meta-transaction id. We can't rely on this existing anymore, so have to bail here. Yev has some plan for getting this all on-chain and in IPFS.

internal/services/fingerprint/consumer.go
```diff
@@ -107,7 +107,7 @@ func (c *Consumer) HandleDeviceFingerprint(ctx context.Context, event *Event) er
 
 	ad, err := models.AftermarketDevices(
 		models.AftermarketDeviceWhere.EthereumAddress.EQ(addr.Bytes()),
-		qm.Load(models.AftermarketDeviceRels.VehicleToken),
+		qm.Load(qm.Rels(models.AftermarketDeviceRels.VehicleToken, models.UserDeviceRels.Claim)),
```
elffjs, line 110: We still use the claim link a few lines down.

internal/controllers/user_devices_controller.go
```diff
@@ -231,75 +231,80 @@ func (udc *UserDevicesController) dbDevicesToDisplay(ctx context.Context, device
...
-			}
-			if nftStatus.Hash.Valid {
-				hash := hexutil.Encode(nftStatus.Hash.Bytes)
-				nft.TxHash = &hash
-			}
 
-			nft.TokenID = d.TokenID.Int(nil)
-			nft.TokenURI = fmt.Sprintf("%s/v1/vehicle/%s", udc.Settings.DeploymentBaseURL, nft.TokenID)
-			if d.OwnerAddress.Valid {
+		if !d.TokenID.IsZero() || d.R.MintRequest != nil {
```
elffjs, line 235: Still display this section if there's no token id. Need to tell the frontend, in particular, how things are going during minting.

internal/controllers/user_devices_controller.go
```diff
@@ -231,75 +231,80 @@ func (udc *UserDevicesController) dbDevicesToDisplay(ctx context.Context, device
...
-				models.NFTPrivilegeWhere.TokenID.EQ(types.Decimal(d.TokenID)),
-				models.NFTPrivilegeWhere.Expiry.GT(time.Now()),
-				models.NFTPrivilegeWhere.ContractAddress.EQ(common.FromHex(udc.Settings.VehicleNFTAddress)),
-			).All(ctx, udc.DBS().Reader)
-			if err != nil {
-				return nil, err
-			}
+				// NFT Privileges
+				udp, err := models.NFTPrivileges(
+					models.NFTPrivilegeWhere.TokenID.EQ(types.Decimal(d.TokenID)),
```
elffjs, line 248: This was actually a bug (well, an inefficiency, we would search for token id 0 for in-progress mints I think) in the old code.

internal/controllers/user_devices_controller.go
```diff
@@ -231,75 +231,80 @@ func (udc *UserDevicesController) dbDevicesToDisplay(ctx context.Context, device
...
-					sdStat.TokenID = sd.TokenID.Int(nil)
-					a := common.BytesToAddress(sd.WalletAddress)
-					sdStat.Address = &a
+					hash := hexutil.Encode(mtr.Hash.Bytes)
+					nft.TxHash = &hash
 				}
 			}
+		}
+
+		if sd := d.R.VehicleTokenSyntheticDevice; sd != nil {
```
elffjs, line 284: Move these out of the big if statement above. No need really.

internal/controllers/nft_controller.go
```diff
@@ -200,6 +200,10 @@ func (nc *NFTController) GetNFTImage(c *fiber.Ctx) error {
 		return opaqueInternalError
 	}
 
+	if !nft.MintRequestID.Valid {
```
elffjs, line 203: Similar fallout of "metatransaction id is filename" decision.

elffjs (commented):
Made some edits. Let me know what you think! I think this is ready to go.

Allyson-English:
> Made some edits. Let me know what you think! I think this is ready to go.

just finished looking through-- those all make sense/ look good to me!

elffjs (approved):

KevinJoiner (approved):

PR #304, opened 2024-04-10
Feature/si 2427 apiv2 consolidate analytics endpoint for dinc service in api

0xdev22:
# Proposed Changes


internal/controllers/user_devices_controller_v2.go
```diff
@@ -0,0 +1,127 @@
...
+
+// GetRange godoc
+// @Description gets the estimated range for a particular user device
+// @Tags        user-devices
+// @Produce     json
+// @Success     200 {object} controllers.DeviceRange
+// @Security    BearerAuth
+// @Param       tokenId path int true "tokenId"
+// @Router      /v2/vehicles/{tokenId}/analytics/range [get]
+func (udc *UserDevicesControllerV2) GetRange(c *fiber.Ctx) error {
```
KevinJoiner, line 49: It looks like most of this is duplicate code from the v1/GetRange. If this is true, can you move it to a single function that both endpoints call?

KevinJoiner (changes requested):

cmd/devices-api/api.go
```diff
@@ -206,6 +199,7 @@ func startWebAPI(logger zerolog.Logger, settings *config.Settings, pdb db.Store,
 	})
 
 	vPriv := app.Group("/v1/vehicle/:tokenID", privilegeAuth)
+	vehiclePrivilegeV2Routes := app.Group("/v2/vehicles/:tokenID", privilegeAuth)
```
Allyson-English, line 202: nit: could shorten this var name to `vPriv2`

Allyson-English (approved):

PR #305, opened 2024-04-10
Add Telemetry/Subscribe command to Tesla Commands

0xdev22:
# Proposed Changes

- Move vehice command to smartcar and tesla client 
- Add telemetry/subscribe to tesla commands


internal/services/models.go
```diff
@@ -152,8 +152,9 @@ type UserDeviceAPIIntegrationsMetadata struct {
 }
 
 type UserDeviceAPIIntegrationsMetadataCommands struct {
-	Enabled []string `json:"enabled,omitempty"`
-	Capable []string `json:"capable,omitempty"`
+	Enabled  []string `json:"enabled,omitempty"`
+	Capable  []string `json:"capable,omitempty"`
+	Disabled []string `json:"disabled,omitempty"`
```
KevinJoiner, line 157: Do we need a new `Disabled` field since we already have `Capable`? If there is a Command in `Capable` but not in `Enabled` this should imply the command is disabled.

KevinJoiner (changes requested):

internal/services/models.go
0xdev22: This makes sense. However, `Capable` is supposed to hold all available commands, splitting things out into a new field called `Disabled` helps avoid having to compute each time what is disabled/enabled which makes the commands easier to work with as we scale or increase the number of available commands.

elffjs: I don’t think we ever ended up using the enabled/disabled stuff—the frontend stores the preference.

internal/services/tesla_fleet_api_service.go
```diff
@@ -155,6 +156,12 @@ func (t *teslaFleetAPIService) WakeUpVehicle(ctx context.Context, token, region
 	return nil
 }
 
+func (t *teslaFleetAPIService) GetAvailableCommands() *UserDeviceAPIIntegrationsMetadataCommands {
+	return &UserDeviceAPIIntegrationsMetadataCommands{
+		Enabled: []string{"doors/unlock", "doors/lock", "trunk/open", "frunk/open", "charge/limit"},
```
KevinJoiner, line 161: Do we also want to set `Capable` field to include `telemetry/subscribe` ?

KevinJoiner (changes requested):

internal/services/tesla_fleet_api_service.go
```diff
@@ -155,6 +156,13 @@ func (t *teslaFleetAPIService) WakeUpVehicle(ctx context.Context, token, region
...
```
Allyson-English, line 161: nit: should we make these command names constants?

KevinJoiner (approved):
🔥

PR #309, opened 2024-04-18
fix log

Allyson-English:
# Proposed Changes


elffjs (approved):
I kinda regret logging this on the GET, but this is more correct than it was.

PR #310, opened 2024-04-18
logging-aftermarketDevicePaired

Allyson-English:
when compiling vehicles to be burned, there were a few that were not associated with a user device id but were paired on chain. seems like the event was never fully processed. adding these logs so that if this happens in the future we can more easily triangulate where the problem is occurring

### DIFF
--- a/internal/services/contracts_events_consumer.go
+++ b/internal/services/contracts_events_consumer.go
@@ -414,23 +414,27 @@ func (c *ContractsEventsConsumer) aftermarketDevicePaired(e *ContractEventData) 
 		return err
 	}
 
-	c.log.Info().Int64("vehicleNode", args.VehicleNode.Int64()).Int64("aftermarketDeviceNode", args.AftermarketDeviceNode.Int64()).Msg("Pairing aftermarket device and vehicle.")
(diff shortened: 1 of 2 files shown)


internal/services/contracts_events_consumer.go
```diff
@@ -414,23 +414,27 @@ func (c *ContractsEventsConsumer) aftermarketDevicePaired(e *ContractEventData)
...
-	c.log.Info().Int64("vehicleNode", args.VehicleNode.Int64()).Int64("aftermarketDeviceNode", args.AftermarketDeviceNode.Int64()).Msg("Pairing aftermarket device and vehicle.")
+	log := c.log.With().Int64("vehicleNode", args.VehicleNode.Int64()).Int64("aftermarketDeviceNode", args.AftermarketDeviceNode.Int64()).Logger()
+	log.Info().Msg("Pairing aftermarket device and vehicle.")
 
 	am, err := models.AftermarketDevices(
 		models.AftermarketDeviceWhere.TokenID.EQ(utils.BigToDecimal(args.AftermarketDeviceNode)),
 	).One(context.TODO(), c.db.DBS().Reader)
 	if err != nil {
+		log.Err(err).Msg("failed to retrieve aftermarket device")
 		return err
```
elffjs, line 425: Could return a wrapped error here. That'll get logged.

KevinJoiner (approved):

PR #311, opened 2024-04-19
Add tesla virtual key status to integration status endpoint

0xdev22:
# Proposed Changes

This introduces the tesla virtual token status into the existing device integration status endpoint. When the tesla integration  status is requested, an API call is also made to fetch if our virtual token has been added to the Tesla device which is requried for telemtry API among other things.

### Impacted Routes
`/user/devices/:userDeviceID/integrations/:integrationID`


internal/controllers/user_integrations_controller.go
```diff
@@ -58,14 +58,94 @@ func (udc *UserDevicesController) GetUserDeviceIntegration(c *fiber.Ctx) error {
...
+	if meta.TeslaRegion == "" {
+		return false, errors.New("missing tesla region")
+	}
+
+	isNewToken, auth, err := udc.ensureTeslaTokenValidty(ctx, integration.RefreshToken.String, integration.AccessExpiresAt.Time)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to get tesla credentials")
+	}
+
+	if isNewToken {
```
elffjs, line 115: Should we update the KTable, too, if we do this? I'm also a little worried about races.

Is the plan to shut off the pollers if telemetry streaming is enabled?

0xdev22: We will not be shutting off the pollers, they will be complementing this actually. The Telemetry server is limited in what it allows us listen for. Also I don't think we need to update the KTables since we won't be using this anywhere else or calling it frequently, this is just to allow the frontend discover what can be activated for an integration type, the next [ticket](https://linear.app/dimo/issue/SI-2558/create-endpoint-to-enable-telemetry-for-an-integration), handles registering a vehicle for telemetry based on what the integration is capable of.

Allyson-English (approved):

internal/controllers/user_integrations_controller.go
```diff
@@ -58,14 +58,94 @@ func (udc *UserDevicesController) GetUserDeviceIntegration(c *fiber.Ctx) error {
...
+		ExternalID: apiIntegration.ExternalID,
+		CreatedAt:  apiIntegration.CreatedAt,
+	}
+	// Handle fetching virtual token status
+	intd, err := udc.DeviceDefSvc.GetIntegrationByID(c.Context(), integrationID)
+	if err != nil {
+		return fiber.NewError(fiber.StatusBadRequest, "invalid integration id")
+	}
+	if intd.Vendor == constants.TeslaVendor {
+		if !apiIntegration.ExternalID.Valid && apiIntegration.AccessToken.Valid && !(apiIntegration.R.UserDevice.VinConfirmed && apiIntegration.R.UserDevice.VinIdentifier.Valid) {
```
elffjs, line 81: What if the Tesla is on the old API (V1)?

internal/controllers/user_integrations_controller.go
```diff
@@ -58,14 +58,94 @@ func (udc *UserDevicesController) GetUserDeviceIntegration(c *fiber.Ctx) error {
...
+
+	resp := GetUserDeviceIntegrationResponse{
+		Status:     apiIntegration.Status,
+		ExternalID: apiIntegration.ExternalID,
+		CreatedAt:  apiIntegration.CreatedAt,
+	}
+	// Handle fetching virtual token status
+	intd, err := udc.DeviceDefSvc.GetIntegrationByID(c.Context(), integrationID)
+	if err != nil {
+		return fiber.NewError(fiber.StatusBadRequest, "invalid integration id")
```
elffjs, line 78: I think we have some helper for converting gRPC errors to Fiber errors?

internal/controllers/user_integrations_controller.go
```diff
@@ -2060,12 +2141,21 @@ type RegisterDeviceIntegrationRequest struct {
...
+
 type GetUserDeviceIntegrationResponse struct {
 	// Status is one of "Pending", "PendingFirstData", "Active", "Failed", "DuplicateIntegration".
 	Status string `json:"status"`
 	// ExternalID is the identifier used by the third party for the device. It may be absent if we
 	// haven't authorized yet.
 	ExternalID null.String `json:"externalId" swaggertype:"string"`
+
+	// Contains further details about tesla integration status
+	Tesla TeslaConnectionStatus `json:"tesla,omitempty"`
```
elffjs, line 2157: If you want to `omitempty`, should this be a `*TeslaConnectionStatus`?

internal/controllers/user_integrations_controller.go
elffjs: This conditional looks a bit odd to me. We're erroring if

```
no external id AND there's an access token
                        AND (the VIN isn't marked unconfirmed OR there's no VIN)
```

It seems like any one of these things (no external id, no access token, no VIN, VIN unconfirmed) is cause for alarm.

internal/services/tesla_fleet_api_service.go
```diff
@@ -164,12 +178,81 @@ func (t *teslaFleetAPIService) GetAvailableCommands() *UserDeviceAPIIntegrations
...
 
-// performTeslaGetRequest a helper function for making http requests, it adds a timeout context and parses error response
-func (t *teslaFleetAPIService) performTeslaGetRequest(ctx context.Context, url, token string) (*http.Response, error) {
+// VirtualTokenConnectionStatus Checks whether vehicles can accept Tesla commands protocol for the partner's public key
+func (t *teslaFleetAPIService) VirtualTokenConnectionStatus(ctx context.Context, token, region, vin string) (bool, error) {
+	baseURL := fmt.Sprintf(t.Settings.TeslaFleetURL, region)
+	url := fmt.Sprintf("%s/api/1/vehicles/fleet_status", baseURL)
+
+	jsonBody := []byte(fmt.Sprintf(`{"vins": ["%s"]}`, vin))
+	body := bytes.NewReader(jsonBody)
```
elffjs, line 187: Can do [`strings.NewReader`](https://pkg.go.dev/strings#NewReader) to avoid getting your hands dirty with `[]byte`.

internal/services/tesla_fleet_api_service.go
```diff
@@ -164,12 +178,81 @@ func (t *teslaFleetAPIService) GetAvailableCommands() *UserDeviceAPIIntegrations
...
 }
 
-// performTeslaGetRequest a helper function for making http requests, it adds a timeout context and parses error response
-func (t *teslaFleetAPIService) performTeslaGetRequest(ctx context.Context, url, token string) (*http.Response, error) {
+// VirtualTokenConnectionStatus Checks whether vehicles can accept Tesla commands protocol for the partner's public key
+func (t *teslaFleetAPIService) VirtualTokenConnectionStatus(ctx context.Context, token, region, vin string) (bool, error) {
+	baseURL := fmt.Sprintf(t.Settings.TeslaFleetURL, region)
+	url := fmt.Sprintf("%s/api/1/vehicles/fleet_status", baseURL)
+
+	jsonBody := []byte(fmt.Sprintf(`{"vins": ["%s"]}`, vin))
```
elffjs, line 186: If we're going to do string manipulation,`fmt.Sprintf(`{"vins": [%q]}`, vin)` feels a little better to me. Minor point.

internal/services/tesla_fleet_api_service.go
```diff
@@ -164,12 +178,81 @@ func (t *teslaFleetAPIService) GetAvailableCommands() *UserDeviceAPIIntegrations
...
+	reqb, err := json.Marshal(reqs)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "POST", "https://auth.tesla.com/oauth2/v3/token", bytes.NewBuffer(reqb))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Content-Type", "application/json")
```
elffjs, line 232: I think the body is to be form-encoded. See [Tesla](https://developer.tesla.com/docs/fleet-api#third-party-token) and [the spec](https://datatracker.ietf.org/doc/html/rfc6749#page-47).

internal/services/tesla_fleet_api_service.go
```diff
@@ -164,12 +178,81 @@ func (t *teslaFleetAPIService) GetAvailableCommands() *UserDeviceAPIIntegrations
...
+		ClientID:     t.Settings.TeslaClientID,
+		RefreshToken: refreshToken,
+	}
+
+	reqb, err := json.Marshal(reqs)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "POST", "https://auth.tesla.com/oauth2/v3/token", bytes.NewBuffer(reqb))
```
elffjs, line 228: Nice to use the context here, I always forget that. Unfortunate to have to hard-code the URL—isn't it a setting? `settings.TeslaTokenURL`?

internal/controllers/user_integrations_controller.go
```diff
@@ -58,14 +58,94 @@ func (udc *UserDevicesController) GetUserDeviceIntegration(c *fiber.Ctx) error {
...
+}
+
+func (udc *UserDevicesController) ensureTeslaTokenValidty(ctx context.Context, refreshToken string, expiry time.Time) (bool, *services.TeslaAuthCodeResponse, error) {
+	const delta = 10 * time.Second
+
+	now := time.Now
+	if now().Before(expiry.Add(-delta)) {
+		return false, nil, nil
+	}
+	authResp, err := udc.teslaFleetAPISvc.RefreshToken(ctx, refreshToken)
```
elffjs, line 143: Do we have to decrypt these tokens first? Refresh and access.

internal/controllers/user_integrations_controller.go
elffjs: We talked a bit—I think we agree that the KTable could end up with a refresh token that is unusable. More to come.

internal/controllers/user_integrations_controller.go
elffjs: `shared.GrpcErrorToFiber`, there it is. I'm not sure it's always applicable, but you might like it here.

internal/controllers/user_integrations_controller.go
```diff
@@ -58,14 +58,119 @@ func (udc *UserDevicesController) GetUserDeviceIntegration(c *fiber.Ctx) error {
...
+	if err != nil {
+		return shared.GrpcErrorToFiber(err, "invalid integration id")
+	}
+
+	if intd.Vendor == constants.TeslaVendor && meta.TeslaAPIVersion == constants.TeslaAPIV2 {
+		if meta.TeslaRegion == "" {
+			return fiber.NewError(fiber.StatusFailedDependency, "missing tesla region")
+		}
+
+		if !apiIntegration.ExternalID.Valid && apiIntegration.AccessToken.Valid && !apiIntegration.R.UserDevice.VinConfirmed && !apiIntegration.R.UserDevice.VinIdentifier.Valid {
```
elffjs, line 93: Maybe these should be ORs?

internal/controllers/user_integrations_controller_test.go
```diff
@@ -1065,3 +1069,158 @@ func (s *UserIntegrationsControllerTestSuite) TestPostTesla_V2_MissingCredential
...
+	accessToken  string
+	refreshToken string
+	expireAt     time.Time
+}
+
+func (m *deviceIntegrationCredentialsMatcher) String() string {
+	return ""
+}
+
+func (m *deviceIntegrationCredentialsMatcher) Matches(x interface{}) bool {
```
elffjs, line 1132: Interesting.

internal/services/tesla_fleet_api_service.go
```diff
@@ -164,12 +179,85 @@ func (t *teslaFleetAPIService) GetAvailableCommands() *UserDeviceAPIIntegrations
...
+	}
+
+	ctxTimeout, cancel := context.WithTimeout(ctx, time.Second*10)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctxTimeout, "POST", t.Settings.TeslaTokenURL, bytes.NewBuffer(reqb))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
```
elffjs, line 237: I don't think changing the content type is enough.

https://stackoverflow.com/questions/19253469/make-a-url-encoded-post-request-using-http-newrequest

internal/services/tesla_fleet_api_service.go
```diff
@@ -164,12 +179,85 @@ func (t *teslaFleetAPIService) GetAvailableCommands() *UserDeviceAPIIntegrations
...
+	// bytes.NewReader(jsonBody)
+
+	resp, err := t.performRequest(ctx, url, token, http.MethodPost, body)
+	if err != nil {
+		return false, fmt.Errorf("could not fetch vehicles for user: %w", err)
+	}
+
+	defer resp.Body.Close()
+
+	var v VirtualTokenConnectionStatusResponse
```
elffjs, line 198: I would move this down to just above line 204.

internal/services/tesla_fleet_api_service.go
```diff
@@ -164,12 +179,75 @@ func (t *teslaFleetAPIService) GetAvailableCommands() *UserDeviceAPIIntegrations
...
-// performTeslaGetRequest a helper function for making http requests, it adds a timeout context and parses error response
-func (t *teslaFleetAPIService) performTeslaGetRequest(ctx context.Context, url, token string) (*http.Response, error) {
+// VirtualTokenConnectionStatus Checks whether vehicles can accept Tesla commands protocol for the partner's public key
+func (t *teslaFleetAPIService) VirtualTokenConnectionStatus(ctx context.Context, token, region, vin string) (bool, error) {
+	baseURL := fmt.Sprintf(t.Settings.TeslaFleetURL, region)
+	url := fmt.Sprintf("%s/api/1/vehicles/fleet_status", baseURL)
+
+	jsonBody := fmt.Sprintf(`{"vins": [%q]}`, vin)
+	body := strings.NewReader(jsonBody)
+	// bytes.NewReader(jsonBody)
```
elffjs, line 189: Delete?

internal/controllers/user_integrations_controller_test.go
```diff
@@ -1065,3 +1069,158 @@ func (s *UserIntegrationsControllerTestSuite) TestPostTesla_V2_MissingCredential
...
+	s.Require().NoError(err)
+
+	s.deviceDefSvc.EXPECT().GetIntegrationByID(gomock.Any(), integration.Id).Return(integration, nil)
+	s.teslaFleetAPISvc.EXPECT().VirtualTokenConnectionStatus(gomock.Any(), accessTk, region, ud.VinIdentifier.String).Return(true, nil)
+
+	request := test.BuildRequest(http.MethodGet, fmt.Sprintf("/user/devices/%s/integrations/%s", ud.ID, integration.Id), "")
+	res, err := s.app.Test(request, 60*1000)
+	s.Assert().NoError(err)
+
+	s.Require().True(res.StatusCode == fiber.StatusOK)
```
elffjs, line 1109: `.Equal`?

internal/controllers/user_integrations_controller_test.go
```diff
@@ -1065,3 +1069,158 @@ func (s *UserIntegrationsControllerTestSuite) TestPostTesla_V2_MissingCredential
...
+	body, _ := io.ReadAll(res.Body)
+
+	defer res.Body.Close()
+
+	actual := GetUserDeviceIntegrationResponse{}
+	s.Assert().NoError(json.Unmarshal(body, &actual))
```
elffjs, line 1115: Do you need to write `.Assert`? I thought all these assertions were tacked onto the suite itself. It's `.Require` for which you have to go the extra mile.

PR #335, opened 2024-05-04
Remove AutoPiTaskService from the AutoPi transaction handler

elffjs:
It wasn't being used anywhere. It's unclear how these tests were passing

# Proposed Changes


internal/services/autopi/integration_test.go
```diff
@@ -47,21 +46,19 @@ const migrationsDirRelPath = "../../../migrations"
 // SetupSuite starts container db
 func (s *IntegrationTestSuite) SetupSuite() {
 	mockCtrl := gomock.NewController(s.T())
-	defer mockCtrl.Finish()
```
elffjs, line 50: Okay, this was the problem—if you call `Finish` before the tests start then you don't assert anything.

In general I think you have a controller per test.

PR #336, opened 2024-05-04
Delete AutoPiTaskService, which isn't used anywhere anymore once /upd…

elffjs:
…ate is gone

I wish this meant that we could kill Redis but alas.

We can, however, get rid of our forked version of taskq.


zer0stars:
I thought we used the autopi task service to set the template?

elffjs:
@JamesReate Giving this a nudge.

JamesReate (approved):

go.mod
```diff
@@ -46,7 +47,6 @@ require (
 	github.com/swaggo/swag v1.16.3
 	github.com/testcontainers/testcontainers-go v0.27.0
 	github.com/tidwall/gjson v1.17.0
-	github.com/vmihailenco/taskq/v3 v3.2.8
```
JamesReate, line 49: yay

JamesReate:
yeah this is good to remove - was only used to apply the update and keep track of that.

PR #340, opened 2024-05-07
SI-2602 Decode and display revert data

elffjs:
The meta-transactor now tries to return Solidity error data if it can along with a ["Failed" transaction message](https://github.com/DIMO-Network/meta-transaction-processor/blob/aa1c3be60cbcb185b527cd466e24117d29b421d6/internal/status/status.go#L183). Here we try to decode this into something human-readable and display it to the client if we can.

The plan is:

1. Take the selector from the error data
2. Look for a match in the ABI
3. See if we defined a template for this error [the mapping file](https://github.com/DIMO-Network/devices-api/pull/340/files#diff-e6082091288094de6f5b73d5f6032d811a2906794403af318ea054164d4a139f)
4. Use the ABI to unpack the arguments into a map
5. Execute the translation template with that map


KevinJoiner:
Should we add the `DimoRegistry.json` and `translation.yaml` files used to generate the code to the repo so it could be regenerated or updated in the future.

internal/services/registry/error_translate.go
```diff
@@ -0,0 +1,71 @@
...
+
+type DimoRegistryErrorDecoder struct {
+	abi         *abi.ABI
+	translation map[string]*template.Template
+}
+
+func NewDimoRegistryErrorDecoder(abi *abi.ABI, translation map[string]string) (*DimoRegistryErrorDecoder, error) {
+	m := make(map[string]*template.Template)
+
+	for errName, message := range translation {
```
elffjs, line 20: Atrocious variable names. Need to go through it.

LorranSutter (approved):

PR #344, opened 2024-05-09
Si 2628 vehicle node minted with device definition

Allyson-English:
# Proposed Changes
- Handle `VehicleNodeMintedWithDeviceDefinition`
- Look up device definition id (ksuid) using the new device definition slug 
- Make sure that device definition and manufacturer token id is correct 
- Look up transaction hash in meta transaction table to make sure we aren't making two copies of vehicles where the mint was initiated by us (as discussed, there is potential for a race here)

### Caveats
- There is potential for this handler to create a new vehicle before the meta transaction processor can update a tx initiated by us with a transaction hash. We decided to go with the method proposed here but expect to update this later on. 

### NOTES:
- this adds a mock device definition service to the `services` package because adding it to `mock_services` created a circular import pattern... is there a better place for this to live/ to test this?


internal/services/contracts_events_consumer.go
```diff
@@ -698,6 +711,136 @@ func (c *ContractsEventsConsumer) aftermarketDeviceAddressReset(e *ContractEvent
 	return err
 }
 
+// vehicleNodeBurned handles the event of the same name from the registry contract.
+func (c *ContractsEventsConsumer) vehicleNodeBurned(e *ContractEventData) error {
```
elffjs, line 715: We're handling this in the `Transfer(address,address,uint256)` now. I like to stick with "basic" events if I can, they're more likely to be emitted.

internal/services/contracts_events_consumer.go
```diff
@@ -698,6 +711,136 @@ func (c *ContractsEventsConsumer) aftermarketDeviceAddressReset(e *ContractEvent
...
+	}
+
+	userIDArgs := dex.IDTokenSubject{
+		UserId: args.Owner.Hex(),
+		ConnId: "web3",
+	}
+	userID, err := proto.Marshal(&userIDArgs)
+	if err != nil {
+		return fmt.Errorf("failed to marshal user id: %w", err)
+	}
```
elffjs, line 793: Shame on me for not doing this before—could we make like `func addressToUserID(common.Address) (string, error)`?

internal/services/contracts_events_consumer.go
```diff
@@ -698,6 +711,136 @@ func (c *ContractsEventsConsumer) aftermarketDeviceAddressReset(e *ContractEvent
...
+	if args.ManufacturerId != big.NewInt(int64(dDef.Make.TokenId)) {
+		return fmt.Errorf("passed manufacturer id %d does not match manufacturer associated with device definition %s", args.ManufacturerId, dDef.DeviceDefinitionId)
+	}
+
+	ud := models.UserDevice{
+		ID:                 ksuid.New().String(),
+		UserID:             base64.RawURLEncoding.EncodeToString(userID),
+		DeviceDefinitionID: dDef.DeviceDefinitionId,
+		OwnerAddress:       null.BytesFrom(args.Owner.Bytes()),
+		TokenID:            types.NewNullDecimal(new(decimal.Big).SetBigMantScale(args.VehicleId, 0)),
```
elffjs, line 813: Can use `dbtypes.NullIntToDecimal`.

internal/services/contracts_events_consumer.go
```diff
@@ -698,6 +711,136 @@ func (c *ContractsEventsConsumer) aftermarketDeviceAddressReset(e *ContractEvent
...
+	}
+
+	if err := ud.Insert(ctx, tx, boil.Infer()); err != nil {
```
elffjs, line 816: I was going to suggest upserting and ignoring on conflict, but we kinda get the same behavior anyway.

internal/services/contracts_events_consumer.go
```diff
@@ -698,6 +711,136 @@ func (c *ContractsEventsConsumer) aftermarketDeviceAddressReset(e *ContractEvent
...
+		return fmt.Errorf("failed to insert new user device: %w", err)
+	}
+
+	c.evtSvc.Emit(&shared.CloudEvent[any]{ //nolint
```
elffjs, line 820: Ah, right, good eye.

internal/services/contracts_events_consumer.go
```diff
@@ -698,6 +711,136 @@ func (c *ContractsEventsConsumer) aftermarketDeviceAddressReset(e *ContractEvent
...
+		return fmt.Errorf("failed to start transaction: %w", err)
+	}
+	defer tx.Rollback() //nolint
+
+	if check, err := models.MetaTransactionRequests(
+		models.MetaTransactionRequestWhere.Hash.EQ(null.BytesFrom(e.TransactionHash.Bytes())),
+	).Exists(ctx, tx); err != nil {
+		return fmt.Errorf("failed to check if vehicle mint event has already been handled: %w", err)
+	} else if check {
+		return fmt.Errorf("vehicle mint event has already been handled. tx hash: %s", e.TransactionHash.String())
```
elffjs, line 783: I'm not sure I would error—that could get a little noisy.

elffjs (approved):
My only complaint is the re-introduction of the burn handler.

internal/services/contracts_events_consumer.go
Allyson-English: ah, snuck in with an out of date git merge!

KevinJoiner (changes requested):

internal/services/device_definition_service_mock.go
```diff
@@ -0,0 +1,248 @@
+// Code generated by MockGen. DO NOT EDIT.
```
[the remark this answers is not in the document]
elffjs: Can you import from other test files? I was under the impression that you couldn't.

internal/services/contracts_events_consumer.go
```diff
@@ -698,10 +702,104 @@ func (c *ContractsEventsConsumer) aftermarketDeviceAddressReset(e *ContractEvent
...
+	dDef, err := c.ddSvc.GetDeviceDefinitionBySlugName(ctx, &ddgrpc.GetDeviceDefinitionBySlugNameRequest{Slug: args.DeviceDefinitionId})
+	if err != nil {
+		return fmt.Errorf("failed to get device definition: %w", err)
+	}
+
+	if dDef.Make.TokenId == 0 {
+		return fmt.Errorf("vehicle make not yet minted: %s", dDef.Make.Name)
+	}
+
+	if args.ManufacturerId.Cmp(big.NewInt(int64(dDef.Make.TokenId))) != 0 {
```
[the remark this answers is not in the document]
elffjs: The manufacturer is more on-chain, to me. On the face of it different manufacturers could have definition ids that clash, but I think James and co have instituted some convention such that you can extract the manufacturer from the definition id. It's like `ford_` at the start or something.

internal/services/contracts_events_consumer.go
```diff
@@ -698,10 +702,104 @@ func (c *ContractsEventsConsumer) aftermarketDeviceAddressReset(e *ContractEvent
...
+	if err != nil {
+		return fmt.Errorf("failed to start transaction: %w", err)
+	}
+	defer tx.Rollback() //nolint
+
+	if check, err := models.MetaTransactionRequests(
+		models.MetaTransactionRequestWhere.Hash.EQ(null.BytesFrom(e.TransactionHash.Bytes())),
+	).Exists(ctx, tx); err != nil {
+		// if we cannot confirm whether event has been handled, log error and then create vehicle
+		log.Err(err).Msg("failed to check if vehicle mint event has already been handled")
```
elffjs, line 729: Interesting to not bail here. Let's see.

internal/services/device_definitions_service.go
```diff
@@ -234,6 +235,18 @@ func (d *deviceDefinitionService) GetIntegrationByVendor(ctx context.Context, ve
 	return integration, nil
 }
 
+func (d *deviceDefinitionService) GetDeviceDefinitionBySlugName(ctx context.Context, in *ddgrpc.GetDeviceDefinitionBySlugNameRequest) (*ddgrpc.GetDeviceDefinitionItemResponse, error) {
+	definitionsClient, conn, err := d.getDeviceDefsGrpcClient()
```
elffjs, line 239: I'm not sure why it creates the client every time, that seems weird.

internal/services/mocks/device_definitions_service_mock.go
```diff
@@ -1,10 +1,5 @@
 // Code generated by MockGen. DO NOT EDIT.
 // Source: device_definitions_service.go
-//
-// Generated by this command:
-//
-//	mockgen -source device_definitions_service.go -destination mocks/device_definitions_service_mock.go
-//
```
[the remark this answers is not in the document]
Allyson-English: I tried updating it and it was still happening.. not sure why. I decided just to not regenerate all of them but... odd!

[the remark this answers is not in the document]
elffjs: Maybe you have the old URL for gomock? It's at go.uber.org/mock/mockgen now

internal/services/device_definition_service_mock.go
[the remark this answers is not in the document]
KevinJoiner: You can't, but if the package using the interface is the one defining it, then other packages would not be importing the mock.

KevinJoiner (approved):

internal/services/device_definition_service_mock.go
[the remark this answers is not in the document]
elffjs: "Interfaces live with consumers" is a very tough battle here.

internal/services/device_definition_service_mock_test.go
```diff
@@ -0,0 +1,248 @@
+// Code generated by MockGen. DO NOT EDIT.
```
elffjs, line 1: Now we have two of these files. I'm not sure this is an improvement. Can we do this the old way for now?

elffjs, line 1: I see. You get an import cycle. I would like to stop dumping everything in `/services` anyway. Let's revisit in a bit.

elffjs:
Let's merge and try it out on dev?

PR #345, opened 2024-05-09
SI-2625-update-mint-evt-in-metatx

Allyson-English:
# Proposed Changes


internal/services/registry/storage.go
```diff
@@ -64,28 +70,28 @@ func (p *proc) Handle(ctx context.Context, data *ceData) error {
...
+	vehicleNodeMintedWithDeviceDefinition := p.ABI.Events["VehicleNodeMintedWithDeviceDefinition"]
 	syntheticDeviceMintedEvent := p.ABI.Events["SyntheticDeviceNodeMinted"]
 
 	switch {
 	case mtr.R.MintRequestUserDevice != nil:
 		for _, logs := range data.Transaction.Logs {
-			if logs.Topics[0] == vehicleMintedEvent.ID {
-				var event contracts.RegistryVehicleNodeMinted
-				err := p.parseLog(&event, vehicleMintedEvent, logs)
+			if logs.Topics[0] == vehicleNodeMintedWithDeviceDefinition.ID {
```
elffjs, line 82: Could we handle both `VehicleNodeMinted` and `VehicleNodeMintedWithDeviceDefinition`? I feel like that might make our upgrade path smoother.

internal/controllers/user_devices_controller.go
```diff
@@ -1580,14 +1569,10 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
-				Owner:            realAddr,
-				AttrInfoPairsVehicle: []contracts.AttributeInfoPair{
-					{Attribute: "Make", Info: deviceMake},
-					{Attribute: "Model", Info: deviceModel},
-					{Attribute: "Year", Info: deviceYear},
-				},
+			return client.MintVehicleAndSdWithDeviceDefinitionSign(requestID, contracts.MintVehicleAndSdWithDdInput{
+				ManufacturerNode:    makeTokenID,
+				Owner:               realAddr,
+				DeviceDefinitionId:  userDevice.DeviceDefinitionID,
```
elffjs, line 1575: I think Lorran is going to merge in a version of this with attributes soon.

elffjs: Can we flag this by environment, actually? It would be nice to still be able to push this to main.

internal/controllers/user_devices_controller.go
```diff
@@ -1417,15 +1412,10 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
-	deviceMake := dd.Make.Name
-	deviceModel := dd.Type.Model
-	deviceYear := strconv.Itoa(int(dd.Type.Year))
-
-	mvs := registry.MintVehicleSign{
-		ManufacturerNode: makeTokenID,
-		Owner:            common.HexToAddress(*user.EthereumAddress),
-		Attributes:       []string{"Make", "Model", "Year"},
-		Infos:            []string{deviceMake, deviceModel, deviceYear},
+	mvs := registry.MintVehicleWithDeviceDefinitionSign{
```
elffjs, line 1415: I think we need a conditional here too.

internal/controllers/user_devices_controller.go
```diff
@@ -1334,15 +1334,10 @@ func (udc *UserDevicesController) GetMintDevice(c *fiber.Ctx) error {
...
```
elffjs, line 1337: And here. It stinks.

elffjs (changes requested):

internal/controllers/user_devices_controller.go
```diff
@@ -1445,15 +1440,45 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
 
+		hash, err = client.Hash(&mvs)
+		if err != nil {
+			return opaqueInternalError
+		}
+	} else {
+		mvs := registry.MintVehicleWithDeviceDefinitionSign{
+			ManufacturerNode:   makeTokenID,
+			Owner:              common.HexToAddress(*user.EthereumAddress),
+			DeviceDefinitionID: userDevice.DeviceDefinitionID,
```
elffjs, line 1466: A realization: this is not the device definition the chain expects. We'll have to convert.

elffjs (changes requested):

internal/controllers/user_devices_controller.go
Allyson-English: oh yeah, great catch!

internal/controllers/user_devices_controller.go
```diff
@@ -1334,15 +1334,21 @@ func (udc *UserDevicesController) GetMintDevice(c *fiber.Ctx) error {
...
-		Owner:            common.HexToAddress(*user.EthereumAddress),
-		Attributes:       []string{"Make", "Model", "Year"},
-		Infos:            []string{deviceMake, deviceModel, deviceYear},
+		return c.JSON(client.GetPayload(&mvs))
+	}
+
+	mvs := registry.MintVehicleWithDeviceDefinitionSign{
+		ManufacturerNode:   makeTokenID,
+		Owner:              common.HexToAddress(*user.EthereumAddress),
+		DeviceDefinitionID: fmt.Sprintf("%s_%s_%d", dd.Make.Name, dd.Type.Model, dd.Type.Year),
```
elffjs, line 1351: I think you need to get it from the DD API. It might already be lying around in that `dd` object. But I know it's more than this.

elffjs (changes requested):

internal/controllers/user_devices_controller.go
elffjs: Here: https://github.com/DIMO-Network/device-definitions-api/blob/50da7d7c322c3b183d35295c6f1cd678ddcfac5b/pkg/grpc/device_definition.proto#L40

elffjs: Maybe we should also check that this is non-empty? A little worried about that.

internal/controllers/user_devices_controller.go
```diff
@@ -1580,30 +1607,69 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
...
-	}, sigBytes)
+	if udc.Settings.IsProduction() {
+		return client.MintVehicleSign(requestID, makeTokenID, realAddr, []contracts.AttributeInfoPair{
+			{Attribute: "Make", Info: dd.Make.Name},
+			{Attribute: "Model", Info: dd.Type.Model},
+			{Attribute: "Year", Info: strconv.Itoa(int(dd.Type.Year))},
+		}, sigBytes)
+	}
+
+	return client.MintVehicleWithDeviceDefinitionSign(requestID, makeTokenID, realAddr, dd.NameSlug, sigBytes)
```
elffjs, line 1672: Isn't there a version of this with attributes now? Do you need to update?

internal/services/registry/client.go
```diff
@@ -212,6 +239,20 @@ func (c *Client) MintVehicleSign(requestID string, manufacturerNode *big.Int, ow
 	return c.sendRequest(requestID, data)
 }
 
+// mintVehicleWithDeviceDefinitionSign(uint256 manufacturerNode, address owner, string deviceDefinitionId, (string,string)[] attrInfo, bytes signature) returns()
```
elffjs, line 242: Yeah, right here, in the sig. I guess that's the downside of `Pack`.

internal/services/registry/client.go
```diff
@@ -191,6 +191,33 @@ func (m *MintVehicleAndSdSign) Message() signer.TypedDataMessage {
...
+	ManufacturerNode   *big.Int
+	Owner              common.Address
+	DeviceDefinitionID string
+}
+
+func (m *MintVehicleWithDeviceDefinitionSign) Name() string {
+	return "MintVehicleWithDeviceDefinitionSign"
+}
+
+func (m *MintVehicleWithDeviceDefinitionSign) Type() []signer.Type {
```
elffjs, line 205: I'm also worried about these types. Is this right?

internal/controllers/user_devices_controller.go
elffjs: https://github.com/DIMO-Network/dimo-identity/blob/811a89081cc89c821cb0b37c2ac6796731af743d/contracts/implementations/nodes/VehicleInternal.sol#L21

internal/services/registry/client.go
```diff
@@ -191,6 +191,33 @@ func (m *MintVehicleAndSdSign) Message() signer.TypedDataMessage {
 	}
 }
 
+// MintVehicleWithDeviceDefinitionSign(uint256 manufacturerNode,address owner, string deviceDefinitionId)
+type MintVehicleWithDeviceDefinitionSign struct {
```
elffjs, line 195: https://github.com/DIMO-Network/dimo-identity/blob/811a89081cc89c821cb0b37c2ac6796731af743d/contracts/implementations/nodes/VehicleInternal.sol#L21

internal/services/registry/client.go
```diff
@@ -75,6 +75,82 @@ func (m *MintVehicleSign) Message() signer.TypedDataMessage {
...
+	}
+}
+
+// MintVehicleAndSdWithDeviceDefinitionSign(uint256 integrationNode)
+type MintVehicleAndSdWithDeviceDefinitionSign struct {
+	IntegrationNode *big.Int
+}
+
+func (m *MintVehicleAndSdWithDeviceDefinitionSign) Name() string {
+	return "MintVehicleAndSdWithDeviceDefinitionSign"
```
elffjs, line 139: I don't think this is a real entity in dimo-identity.

elffjs: Like in both cases, this is what's used https://github.com/DIMO-Network/dimo-identity/blob/811a89081cc89c821cb0b37c2ac6796731af743d/contracts/implementations/MultipleMinter.sol#L28

Allyson-English: isn't that this (from the go file generated with the abi): https://github.com/DIMO-Network/devices-api/blob/main/internal/contracts/registry.go#L1942

elffjs: That's a function, not an EIP-712 payload. The payloads always look like what I linked.

PR #348, opened 2024-05-10
SI-2659-store-nft-ipfs

Allyson-English:
# Proposed Changes
update vehicle minting so that a user is required to upload an image before calling mint function (note that this is the opposite of what happens right now, where we require that a vehicle be minted before calling the update image func)

### Notes
- This piles on to the monstrosity that is the `userDeviceController`
- This means that any open vehicle mints will not have the vehicle nft image stored as an attribute but vehicles minted through our app will... are we okay with this lack of parity? 
- Need to chat with mobile about how this will impact their flow


internal/services/ipfs/ipfs.go
```diff
@@ -0,0 +1,79 @@
...
+func NewIPFSLoader(settings *config.Settings) *IPFS {
+	return &IPFS{
+		client: &http.Client{
+			Timeout: 5 * time.Second,
+		},
+		url: settings.IPFSURL,
+	}
+}
+
+func (i *IPFS) UploadImage(img string) (string, error) {
```
elffjs, line 35: Could we take in a context here?

internal/services/ipfs/ipfs.go
```diff
@@ -0,0 +1,79 @@
...
+	var respb ipfsResponse
+	imageData := strings.TrimPrefix(img, "data:image/png;base64,")
+	image, err := base64.StdEncoding.DecodeString(imageData)
+	if err != nil {
+		return respb.CID, err
```
elffjs, line 40: I would just do `return "", err`. The way it is is a little confusing to me.

internal/services/ipfs/ipfs.go
elffjs, line 35: Could we take in a context here and use it for the request?

internal/services/ipfs/ipfs.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	if len(image) == 0 {
+		return respb.CID, fmt.Errorf("Empty image field.")
```
elffjs, line 44: Piling onto the linter: we can do `errors.New` if there's nothing to interpolate.

internal/services/ipfs/ipfs.go
```diff
@@ -0,0 +1,79 @@
...
+	return &IPFS{
+		client: &http.Client{
+			Timeout: 5 * time.Second,
+		},
+		url: settings.IPFSURL,
+	}
+}
+
+func (i *IPFS) UploadImage(img string) (string, error) {
+	var respb ipfsResponse
```
elffjs, line 36: I would move this down to where it's used.

migrations/20240510181122_store_nft_img_ipfs.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT 'up SQL query';
+-- +goose StatementEnd
+
+SET search_path = devices_api, public;
+ALTER TABLE user_devices ADD COLUMN ipfs_cid VARCHAR;
```
elffjs, line 7: This looks malformed.

elffjs, line 7: Maybe name it `image_cid` or `ipfs_image_cid`? I could see it being a little ambiguous.

internal/services/ipfs/ipfs.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	if !respb.Success {
+		return respb.CID, fmt.Errorf("failed to upload image to IPFS")
+	}
+
+	return respb.CID, nil
+}
+
+func (i *IPFS) EncodeImage(path string) (string, error) {
```
elffjs, line 68: For testing, I assume?

internal/controllers/user_devices_controller.go
```diff
@@ -1578,6 +1582,15 @@ func (udc *UserDevicesController) UpdateNFTImage(c *fiber.Ctx) error {
...
 	}
 
+	cid, err := udc.ipfsSvc.UploadImage(nid.ImageData)
+	if err != nil {
+		udc.log.Err(err).Msg("failed to upload NFT image to IPFS")
+	}
+	userDevice.IpfsCid = null.StringFrom(cid)
+	if _, err := userDevice.Update(c.Context(), udc.DBS().Writer, boil.Whitelist(models.UserDeviceColumns.IpfsCid)); err != nil {
+		udc.log.Err(err).Str("userDeviceID", userDeviceID).Str("cid", cid).Msg("failed to store IPFS CID for vehicle")
+	}
```
elffjs, line 1592: I think this should all be in an else block.

elffjs (commented):
Premature comments.

internal/services/ipfs/ipfs.go
```diff
@@ -0,0 +1,79 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/DIMO-Network/devices-api/internal/config"
+)
+
```
elffjs, line 15: A good case for constants with `image/png` and the prefix.

internal/controllers/user_devices_controller.go
elffjs: I assume we'll do this in the vehicle mint too?

migrations/20240510181122_store_nft_img_ipfs.sql
Allyson-English: what do you mean?

elffjs: StatementEnd before the SQL starts?

internal/services/ipfs/ipfs_test.go
```diff
@@ -0,0 +1,59 @@
...
...ZwaVDIhe9Ov6mY0/5ImqBSrXFKzZN5K39/dQYeUTxVgowEkfCGKm4diaPZLpot5TFL2mEqsRBnlEzE3odQYT3wGTjRPuHiW4ndMRTStmrGcJvCKxmu5Qe/S/Eke4MaIQvEdriNC8T5GBZ+zydmVZTmTm3HvPW8eePPY4ODnoKyVW6qa/afmfadOFoQib9zS5GAaueG0CRAWZthKrBrHoiu6LhBtJbomOGM5VmED81HR4gzauBEiZ8owlQtf9sU4KF4dVN0wCeDAOGCZ3M9DgY2XpL64VjgpL3/g3XffezMcDoHNb537m4bVjh9F+9IsOMdzvT++evFdOo9r3omyoQziPck8knlRoMSotAvWKjuPV0HkjNuJPS9VAo9NIGNormUGJlpiMbwmlDIeTtrxftYXcs/+uauk9I/vHHlHjUYjYKWN3w+teHDILznTCPXEWWWz1ywnU2hNxlfsIgdTDuU0vZBV6MhkOKngG9TrDCWjKrxT2wVGcw8xnlaymdGskexLKZKin7ATwfayqpGko/KDhqY5+5ctXRx7kmaEQj5fNMGySlHk7FnLrIrfo6bxPCL9Ufib3OYxgQOe4jxyLXh82YDx0dDgcaxpOsumCkaoM9F35Ux4fzqW3+v1sPOpsIKKqtm6aJUMHjI9ly4im81OyiCgxbOfaOThzPwNNWX7YLOeoj6soIyOE6aKr2DYTKDrIH2R6olk412KOp4eH04PRLrTnySHcn+JjaSO2b328FiIsm4KaxLoRPMWmAYxcxZe4yAc7H1UNI4QaMrUUjoWTxTJa51+zd8mcbxK0EnT7EoEkA8VFApgupg7/X7ziQXTS39XGUy9+t6RD/X+3s6pYLwK9DpeeI3jgI0CWLYLUrYDsvFP2Jxh2KQklZQYdZc4ZEmBza7DaqUmwIs9HT0EC0yDZKuMXE6kriMhnqzFwNAtSFhoPi2EtDrlpddvbu5+rrur7/Kpjz6kOMxNCejX5/k4a+bNHGvocKbnacJR6FmOTskJuohFpoxvpxrF6EAID3TRBUN08+CaFCQQAc7Eu6M31b71XNCT2LvvTy/q/H9P3xhoEWBxGw1NxRevOd9Nen+i7YyP3JOCx0uQYT6/Giti7rJ6i/PX2ySjf+87bx/4WqBTK0bfwuK2lsTfGL69/s/L/vrKyz032vsvtttAHOpjPP8AAAAASUVORK5CYII="
```
elffjs, line 20: Does it need to be so large xdd

migrations/20240510181122_store_nft_img_ipfs.sql
elffjs: I will just fix it.

internal/services/ipfs/ipfs.go
```diff
@@ -0,0 +1,79 @@
...
+		},
+		url: settings.IPFSURL,
+	}
+}
+
+func (i *IPFS) UploadImage(ctx context.Context, img string) (string, error) {
+	imageData := strings.TrimPrefix(img, imagePrefix)
+	image, err := base64.StdEncoding.DecodeString(imageData)
+	if err != nil {
+		return "", err
```
elffjs, line 45: Could we wrap this error too?

cmd/devices-api/api.go
```diff
@@ -92,7 +93,7 @@ func startWebAPI(logger zerolog.Logger, settings *config.Settings, pdb db.Store,
 	ddIntSvc := services.NewDeviceDefinitionIntegrationService(pdb.DBS, settings)
 	ddSvc := services.NewDeviceDefinitionService(pdb.DBS, &logger, settings)
 	ddaSvc := services.NewDeviceDataService(settings.DeviceDataGRPCAddr, &logger)
-
+	ipfsSvc := ipfs.NewIPFSLoader(settings)
```
elffjs, line 96: I think the Go thing is to not repeat the name, so `ipfs.NewLoader`.

internal/controllers/user_devices_controller.go
```diff
@@ -1453,6 +1457,12 @@ func (udc *UserDevicesController) PostMintDevice(c *fiber.Ctx) error {
 		return opaqueInternalError
 	}
 
+	cid, err := udc.ipfsSvc.UploadImage(c.Context(), imageData)
+	if err != nil {
+		udc.log.Err(err).Msg("failed to upload NFT image to IPFS while minting")
+	}
+	userDevice.IpfsImageCid = null.StringFrom(cid)
```
elffjs, line 1464: Do you still want to do this if the above fails?

internal/services/ipfs/ipfs.go
```diff
@@ -0,0 +1,79 @@
...
+	}
```
elffjs, line 46: Can we wrap the unwrapped here?

internal/controllers/user_devices_controller.go
```diff
@@ -1619,6 +1629,16 @@ func (udc *UserDevicesController) UpdateNFTImage(c *fiber.Ctx) error {
 		return fiber.NewError(fiber.StatusBadRequest, "Couldn't parse request body.")
 	}
 
+	cid, err := udc.ipfsSvc.UploadImage(c.Context(), nid.ImageData)
+	if err != nil {
+		udc.log.Err(err).Msg("failed to upload NFT image to IPFS")
```
elffjs, line 1634: I suppose at some point we'll start failing on this.

internal/services/ipfs/ipfs.go
```diff
@@ -0,0 +1,79 @@
...
+
+	reader := bytes.NewReader(image)
+	req, err := http.NewRequest(http.MethodPost, i.url, reader)
+	if err != nil {
+		return "", fmt.Errorf("failed to create image upload req: %v", err)
+	}
+
+	req.Header.Set("Content-Type", contentTypeHeader)
+	resp, err := i.client.Do(req.WithContext(ctx))
```
elffjs, line 59: Cool, I always forget about `WithContext`.

elffjs (changes requested):

internal/controllers/user_devices_controller.go
Allyson-English: oo nope.. I thought that having an empty string would make valid = false, just tested it  to double check and turns out that is not the case

cmd/devices-api/api.go
```diff
@@ -92,7 +93,7 @@ func startWebAPI(logger zerolog.Logger, settings *config.Settings, pdb db.Store,
 	ddIntSvc := services.NewDeviceDefinitionIntegrationService(pdb.DBS, settings)
 	ddSvc := services.NewDeviceDefinitionService(pdb.DBS, &logger, settings)
 	ddaSvc := services.NewDeviceDataService(settings.DeviceDataGRPCAddr, &logger)
-
+	ipfsSvc := ipfs.NewLoader(settings)
```
elffjs, line 96: Also, does "loader" make sense to you? I tend to assume that loaders are for loading-out-of.

migrations/20240510181122_store_nft_img_ipfs.sql
Allyson-English: oh shoot, sorry I was looking at the sql itself, I never really pay attention to the goose stuff, thanks for fixing that. I would have, I was just confused.

cmd/devices-api/api.go
Allyson-English: updated this to `NewGateway` that has an `UploadImage` method... maybe thats worse. open to other ideas!

elffjs: That might be a bit too broad.

internal/services/ipfs/ipfs.go
```diff
@@ -0,0 +1,104 @@
...
+		},
+		url: settings.IPFSURL,
+	}
+}
+
+func (i *IPFS) UploadImage(ctx context.Context, img string) (string, error) {
+	imageData := strings.TrimPrefix(img, imagePrefix)
+	image, err := base64.StdEncoding.DecodeString(imageData)
+	if err != nil {
+		return "", fmt.Errorf("failed to decode image: %v", err)
```
elffjs, line 46:
```suggestion
		return "", fmt.Errorf("failed to decode image: %w", err)
```
Wrap, wrap, wrap.

internal/services/ipfs/ipfs.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	if len(image) == 0 {
+		return "", errors.New("empty image field")
+	}
+
+	reader := bytes.NewReader(image)
+	req, err := http.NewRequest(http.MethodPost, i.url, reader)
+	if err != nil {
+		return "", fmt.Errorf("failed to create image upload req: %v", err)
```
elffjs, line 56: Same thing.

internal/services/ipfs/ipfs.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	req.Header.Set("Content-Type", contentTypeHeader)
+	resp, err := i.client.Do(req.WithContext(ctx))
+	if err != nil {
+		return "", fmt.Errorf("IPFS post request failed: %v", err)
```
elffjs, line 62: This is getting repetitive—I'll change these.

elffjs (changes requested):

internal/config/settings.go
```diff
@@ -91,6 +91,7 @@ type Settings struct {
 	TeslaTelemetryHostName            string `yaml:"TESLA_TELEMETRY_HOST_NAME"`
 	TeslaTelemetryPort                int    `yaml:"TESLA_TELEMETRY_PORT"`
 	TeslaTelemetryCACertificate       string `yaml:"TESLA_TELEMETRY_CA_CERTIFICATE"`
+	IPFSURL                           string `yaml:"IPFS_URL"`
```
elffjs, line 94: I feel we had better populate this in the values files before we merge. What should the values be?

elffjs (changes requested):

charts/devices-api/values-prod.yaml
```diff
@@ -8,12 +8,12 @@ cronJob:
   schedule: 0 0 1 * *
 env:
   ENVIRONMENT: prod
-  PORT: '8080'
+  PORT: "8080"
```
elffjs, line 11: Whatever is doing these mass reformats, can it be turned off?

internal/services/ipfs/ipfs_test.go
```diff
@@ -0,0 +1,50 @@
...
+		IPFSURL: "https://assets.dev.dimo.xyz/ipfs",
+	})
+
+	pdb, _ := test.StartContainerDatabase(context.Background(), t, "../../../migrations")
+
+	cid, err := ipfs.UploadImage(ctx, img)
+	assert.NoError(t, err)
+	assert.Equal(t, "Qme23PqtDXmeyETzG3W3sy3ZWTjF2ZQGJWrCG5svtFq8aB", cid)
+
+	imgB, err := ipfs.FetchImage(ctx, cid)
```
elffjs, line 33: Does this test depend on this API being up? That's kinda scary to me.

internal/services/ipfs/ipfs_test.go
```diff
@@ -0,0 +1,50 @@
...
+	assert.Equal(t, base64Encoding, img)
+
+	ud := models.UserDevice{
+		ID:                 ksuid.New().String(),
+		UserID:             ksuid.New().String(),
+		DeviceDefinitionID: ksuid.New().String(),
+		IpfsImageCid:       null.StringFrom(cid),
+	}
+
+	if err := ud.Insert(ctx, pdb.DBS().Writer, boil.Infer()); err != nil {
```
elffjs, line 46: Is the database being useful here?

elffjs:
Update your SQLBoiler!

internal/controllers/nft_controller.go
```diff
@@ -115,10 +118,15 @@ func (nc *NFTController) GetNFTMetadata(c *fiber.Ctx) error {
 		name = description
 	}
 
+	imageURI := fmt.Sprintf("%s/v1/vehicle/%s/image", nc.Settings.DeploymentBaseURL, ti)
+	if !nc.Settings.IsProduction() && ud.IpfsImageCid.Valid {
+		imageURI = fmt.Sprintf("%s/%s", ipfs.Scheme, ud.IpfsImageCid.String)
```
elffjs, line 123: This would result in `ipfs:///CiDwhatvEr`, yeah? Maybe a function that goes `cid` → URL just to centralize this?

PR #351, opened 2024-05-22
SI-1844: Synthetic Device Data is Signed

zakharenkodmytro:
# Proposed Changes
We want to expose the wallet child number through proto so that we can later use it to sign the payload.


JamesReate:
from discussion, you'll need to implement getting the value here: https://github.com/DIMO-Network/devices-api/blob/2fa6fa9f80d259b2dd740dfb1d6afc9637f0ebcd/internal/rpc/user_devices.go#L79

and probably the get user device by token id as well etc

internal/rpc/user_devices.go
```diff
@@ -451,9 +451,11 @@ func (s *userDeviceRPCServer) deviceModelToAPI(ud *models.UserDevice) *pb.UserDe
 		if sd := ud.R.VehicleTokenSyntheticDevice; sd != nil && !sd.TokenID.IsZero() {
 			stk, _ := sd.TokenID.Uint64()
 			iTkID, _ := sd.IntegrationTokenID.Uint64()
+			wc := sd.WalletChildNumber
```
zakharenkodmytro, line 454: @JamesReate , is it all that is required for mapping?

JamesReate: looks like it yeah

JamesReate (approved):

PR #352, opened 2024-05-29
SI-2689 Start software polling jobs after minting with on-chain data

elffjs:
# Proposed Changes


cmd/devices-api/main.go
```diff
@@ -91,7 +91,6 @@ func main() {
 		subcommands.Register(&remakeUserDeviceTokenTableCmd{logger: logger, settings: settings, pdb: pdb, container: deps}, "device integrations")
 		subcommands.Register(&remakeFenceTopicCmd{logger: logger, settings: settings, pdb: pdb}, "device integrations")
 		subcommands.Register(&remakeDeviceDefinitionTopicsCmd{logger: logger, settings: settings, pdb: pdb, ddSvc: deps.getDeviceDefinitionService()}, "device integrations")
-		subcommands.Register(&startSmartcarFromRefreshCmd{logger: logger, settings: settings, pdb: pdb, ddSvc: deps.getDeviceDefinitionService()}, "device integrations")
```
zakharenkodmytro, line 94: I assume this is just refactoring not related to the current story backed by PR?

elffjs: No, this is a very old task that we don't use anymore—Smartcar doesn't give us refresh tokens anymore if things go wrong—and I don't want to upgrade it to use this stuff. I think it's kinda related.

internal/services/tesla_task_service.go
```diff
@@ -9,20 +9,21 @@ import (
...
 	"github.com/Shopify/sarama"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/segmentio/ksuid"
 )
 
 //go:generate mockgen -source tesla_task_service.go -destination mocks/tesla_task_service_mock.go
 
 type TeslaTaskService interface {
-	StartPoll(vehicle *TeslaVehicle, udai *models.UserDeviceAPIIntegration, apiVersion int, region string) error
+	StartPoll(udai *models.UserDeviceAPIIntegration, sd *models.SyntheticDevice) error
```
zakharenkodmytro, line 21: A bit confusing "StartPoll" name it looks more like produce the message instead of the poll.

elffjs: It sends out the task description to be executed.

zakharenkodmytro: Maybe just `StartTask` and provide the task's type parameter. Though, I'm not sure if we will add more tasks to TeslaTaskService in the future.

zakharenkodmytro:
@elffjs , it seems that this is self-sufficient PR on the devices-api side. To reiterate for myself:
1. We always include the `SyntheticDevice` to the StartPoll credentials message, which would be stored in `table.task.credential` table
2. This message be produced on every mint device operation. In that case, we do not care about burn device operation at all, I guess.
3. The `oem-task-processor` could open the view to the `table.task.credential` table, which now stores the Synthetic device information and is keyed by TaskId.
4.  The `oem-task-processor` would check the childwallet and if it is presented then go ahead and sign the payload, otherwise drop task
5. All old records in the `table.task.credential` table should be updated

elffjs:
OEM already reads in the entire credential topic and keeps its view updated. From its perspective everything is very simple now. The big lift is updating all the existing records without losing anything, yes.

zakharenkodmytro (approved):

zakharenkodmytro:
I'm just curious: When the device is burned and minted again, will `userDeviceId` change as well? So, when the task arrives at OEM, would it have `userDeviceId` associated with the newly minted vehicle?

elffjs:
If the _vehicle_ is burned, the user device id is gone forever.

If the synthetic device is burned, the vehicle and the user device id remain the same.

zakharenkodmytro:
> OEM already reads in the entire credential topic and keeps its view updated. From its perspective everything is very simple now. The big lift is updating all the existing records without losing anything, yes.

Right now `oem` only stores access token I guess, so we need to store `WalletChildNumber` I guess as well. I'll create PR for this or modify one of the existing ones.

elffjs:
I will update OEM with the shared model in a second.

zakharenkodmytro:
> I will update OEM with the shared model in a second.

Great, it makes things much easier

zakharenkodmytro:
Created corresponding PR on oem side https://github.com/DIMO-Network/oem-task-processor/pull/20

JamesReate (approved):

PR #356, opened 2024-06-13
SI-2726 Emit privacy zones by vehicle id

elffjs:
We already emit privacy fence updates to `table.device.privacyfence`, keyed by user device id, when a fence is modified or when a fence is attached to or detached from a vehicle. Here we also emit the same messages to `table.device.privacyfence.v2`, keyed by vehicle token id.

Asides

* You can't attach or detach a fence from a vehicle that isn't minted.
* Fixes a [long-standing bug](https://github.com/DIMO-Network/devices-api/pull/356/files#diff-363f4238e0f752e79e0a6f611107a294ae1e44b37ee93c2d1705216563c96c66L332): anyone could attach a privacy fence they own to any vehicle, even one they did not own.

Downsides

* If you have an un-minted vehicle that already has fences attached (under the old controller), and you later mint it, we don't emit updates.

I want to clean this file up more, but this PR is already large.

### DIFF
--- a/charts/devices-api/values.yaml
+++ b/charts/devices-api/values.yaml
@@ -226,6 +226,14 @@ kafka:
         min.compaction.lag.ms: '3600000'
         min.cleanable.dirty.ratio: '0.01'
         delete.retention.ms: '3600000'
+    - name: table.device.privacyfence.v2
+      config:
+        segment.ms: '3600000'
+        compression.type: producer
+        cleanup.policy: compact
+        min.compaction.lag.ms: '3600000'
+        min.cleanable.dirty.ratio: '0.01'
+        delete.retention.ms: '3600000'
     - name: devices-vin-fraud-table
       config:
         segment.ms: '3600000'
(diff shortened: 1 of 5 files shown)


charts/devices-api/values.yaml
```diff
@@ -120,6 +120,7 @@ env:
     -----END CERTIFICATE-----
   IPFS_URL: https://assets.dev.dimo.xyz/ipfs
   SD_INFO_TOPIC: table.task.synthetic.fill
+  PRIVACY_FENCE_TOPIC_V2: table.device.privacyfence.v2
```
zakharenkodmytro, line 123: looks like you has another pr for table creation in kafka

internal/controllers/geofences_controller.go
```diff
@@ -329,21 +329,24 @@ func (g *GeofencesController) Update(c *fiber.Ctx) error {
 	if err != nil {
 		return errors.Wrap(err, "error updating geofence")
 	}
-	for _, uID := range update.UserDeviceIDs {
-		affectedDeviceIDs.Add(uID)
+
+	uds, err := g.createDeviceList(c.Context(), tx, userID, update.UserDeviceIDs)
```
elffjs, line 333: Eh this has problems actually. Will fix it up.

charts/devices-api/values.yaml
elffjs: I was playing with too many branches. It's in here now.

zakharenkodmytro (approved):

elffjs:
Going to pare this back a bit. I got too carried away fixing stuff.

PR #358, opened 2024-06-14
SI-2726 NFT privacy zones backfill

elffjs:
Add a subcommand `devices-api populate-privacy-topic-v2`. Sends out privacy zones from #356 keyed by token id.


zakharenkodmytro (approved):
lgtm

PR #366, opened 2024-07-10
[SM-857] change to tableland id

Asfiroth:
# Proposed Changes
- Change lookup for tableland id


migrations/20240703182015_user_device_tableland_id.sql
```diff
@@ -0,0 +1,12 @@
+-- +goose Up
+-- +goose StatementBegin
+SET search_path = devices_api, public;
+ALTER TABLE user_devices ADD COLUMN definition_id text not null default '';
```
JamesReate, line 4: let's make it nullable to start, that way don't need to default to a blank string for now. Or was your thought to backfill and then save this step? We'd probably want to remove the default '' after doing that, like either way i think we'll need a migration after backfilling

internal/services/user_device_service.go
```diff
@@ -101,7 +101,7 @@ func (uds *userDeviceService) CreateUserDevice(ctx context.Context, deviceDefID,
 	ud := models.UserDevice{
 		ID:                 userDeviceID,
 		UserID:             userID,
-		DeviceDefinitionID: dd.DeviceDefinitionId,
+		DeviceDefinitionID: dd.NameSlug,
```
JamesReate, line 104: the legacy `DeviceDefinitionID` should still be the ksuid legacy id. It would be the new column that would get set. Realize `DefinitionID` is confusing for the new name, but the idea is to remove the DeviceDefinitionID eventually so...

internal/controllers/user_devices_controller.go
```diff
@@ -596,20 +600,26 @@ func (udc *UserDevicesController) RegisterDeviceForUserFromVIN(c *fiber.Ctx) err
...
+
 	var udFull *UserDeviceFull
+
 	if existingUD != nil {
 		if existingUD.UserID != userID {
 			return fiber.NewError(fiber.StatusConflict, "VIN already exists for a different user: "+reg.VIN)
 		}
-		deviceDefinitionID = existingUD.DeviceDefinitionID
+
+		deviceDefinitionID = existingUD.DefinitionID
```
JamesReate, line 615: not sure on doing this?

internal/controllers/devices_controller.go
```diff
@@ -19,7 +19,7 @@ func NewDeviceDefinitionFromGRPC(dd *grpc.GetDeviceDefinitionItemResponse) (serv
 		}
 	}
 	rp := services.DeviceDefinition{
-		DeviceDefinitionID:     dd.DeviceDefinitionId,
+		DeviceDefinitionID:     dd.NameSlug,
```
JamesReate, line 22: same here, see below

cmd/devices-api/sync_user_device_definition.go
```diff
@@ -0,0 +1,91 @@
...
+		return subcommands.ExitFailure
+	}
+
+	return subcommands.ExitSuccess
+}
+
+func (s *syncUserDeviceDeviceDefinitionCmd) syncUserDevicesWithTableland(ctx context.Context) error {
+	dbs := s.pdb.DBS()
+
+	userDevices, err := models.UserDevices().All(ctx, dbs.Reader)
```
JamesReate, line 54: i think this will crash b/c we have more than 65537 etc records. Will need paging, maybe do 5k records at a time. You could also stream the data but that may be harder cause you have to use native lower level sql lib

cmd/devices-api/sync_user_device_definition.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func ProcessDeviceDefinitions(ctx context.Context, dbs *db.ReaderWriter, deviceDefinitions []*ddgrpc.GetDeviceDefinitionItemResponse) error {
+	cursor := ""
+	hasMore := true
+
+	for hasMore {
+		userDevices, err := models.UserDevices(
+			models.UserDeviceWhere.DefinitionID.IsNull(),
+			models.UserDeviceWhere.ID.GT(cursor),
```
JamesReate, line 87: i guess it helps ksuid's can be ordered alphabetically

internal/controllers/device_data_controller.go
```diff
@@ -151,9 +151,9 @@ func (udc *UserDevicesController) QueryDeviceErrorCodes(c *fiber.Ctx) error {
 		return err
 	}
 
-	dd, err := udc.DeviceDefSvc.GetDeviceDefinitionByID(c.Context(), ud.DeviceDefinitionID)
+	dd, err := udc.DeviceDefSvc.GetDeviceDefinitionByID(c.Context(), ud.DefinitionID.String)
```
JamesReate, line 154: this looks like changing the underlying way the GRPC endpoint works as to what key it uses... Is there an accompanying PR for this change? I don't know how many places use that endpoint but I think it is a fair amount, I would be more comfortable with doing a v2 of it, transitioning searching in all repos, then moving it over.

JamesReate (approved):

elffjs:
What’s going on with this one?

JamesReate:
superseded by #402

PR #369, opened 2024-07-16
Drop verifiable_credentials table

elffjs:
We haven't been creating VCs in a while, but some stuff is still hanging around:

* Database
   * Recreates the `vehicle_nfts` view so that `claim_id` is always null 
   * Drop column `claim_id` on `user_devices`
   * Drop table `verifiable_credentials`
* API
   * Remove the route that gets VCs
   * Remove VCs in the `/user/devices` responses
   * VC field deprecated and always null in gRPC responses


elffjs:
@JamesReate If I merge this, will it break admin?

elffjs:
Well I'll just merge it and if admin breaks then we'll fix it.

PR #391, opened 2024-09-27
Feature/as 3033

elffjs:
# Proposed Changes

### DIFF
--- a/cmd/devices-api/api.go
+++ b/cmd/devices-api/api.go
@@ -198,6 +198,7 @@ func startWebAPI(logger zerolog.Logger, settings *config.Settings, pdb db.Store,
 	vehicleAddr := common.HexToAddress(settings.VehicleNFTAddress)
 
 	// vehicle command privileges
+	vPriv.Patch("/vin", privTokenWare.OneOf(vehicleAddr, []privileges.Privilege{privileges.VehicleCommands}), userDeviceController.UpdateVIN)
 	vPriv.Post("/commands/doors/unlock", privTokenWare.OneOf(vehicleAddr, []privileges.Privilege{privileges.VehicleCommands}), nftController.UnlockDoors)
 	vPriv.Post("/commands/doors/lock", privTokenWare.OneOf(vehicleAddr, []privileges.Privilege{privileges.VehicleCommands}), nftController.LockDoors)
 	vPriv.Post("/commands/trunk/open", privTokenWare.OneOf(vehicleAddr, []privileges.Privilege{privileges.VehicleCommands}), nftController.OpenTrunk)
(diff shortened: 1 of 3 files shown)


elffjs:
Okay, we gotta fix the test

cmd/devices-api/api.go
```diff
@@ -198,6 +198,7 @@ func startWebAPI(logger zerolog.Logger, settings *config.Settings, pdb db.Store,
 	vehicleAddr := common.HexToAddress(settings.VehicleNFTAddress)
 
 	// vehicle command privileges
+	vPriv.Patch("/vin", privTokenWare.OneOf(vehicleAddr, []privileges.Privilege{privileges.VehicleCommands}), userDeviceController.UpdateVIN)
```
JamesReate, line 201: should we create a new privilege for updating the VIN? Like what if the user doesn't have the VehicleCommands privilege, or is `provides access to all commands that can be sent to the device.` one of the default privileges you get when pairing

elffjs: The user has all privileges on his own NFT.

internal/controllers/nft_controller.go
```diff
@@ -437,6 +438,133 @@ func (nc *NFTController) GetSyntheticDeviceNFTMetadata(c *fiber.Ctx) error {
...
+func validVINChar(r rune) bool {
+	return 'A' <= r && r <= 'Z' || '0' <= r && r <= '9'
+}
+
+// UpdateVIN godoc
+// @Description updates the VIN on the user device record
+// @Tags        user-devices
+// @Produce     json
+// @Accept      json
+// @Param      	tokenId path int true "token id"
```
JamesReate, line 450:
```suggestion
// @Param      	tokenId path int true "token id"
// @Param  vinReq body UpdateVINReq true  "Update VIN request"
```

elffjs: Yeah that's good. I'll fix that.

JamesReate (approved):

internal/controllers/nft_controller.go
```diff
@@ -437,6 +438,133 @@ func (nc *NFTController) GetSyntheticDeviceNFTMetadata(c *fiber.Ctx) error {
...
+	).One(c.Context(), tx)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return fiber.NewError(fiber.StatusNotFound, fmt.Sprintf("Vehicle NFT %d not found.", tokenID))
+		}
+		logger.Err(err).Msg("Failed to search for device.")
+		return opaqueInternalError
+	}
+
+	if userDevice.VinConfirmed {
```
zer0stars, line 496: @mateuszpohl

PR #394, opened 2024-10-12
implement move all force script

JamesReate:
# Proposed Changes
- ap script to move remaining laggard devices to tmpl 130
- this may set off alarm on certain cars that are finicky


JamesReate:
ran go generate locally exactly as script runs but still getting dirty files

PR #399, opened 2024-10-20
Adapt to new AWS SDK conventions: no more global region

elffjs:
# Proposed Changes


JamesReate:
i like that this solves that nagging linter error

elffjs:
I wonder if it's gonna work.

elffjs:
Let's see.

elffjs:
Bold of you to tag.

PR #410, opened 2024-11-04
Sm 1245 change where integration status is updated from

elffjs:
# Proposed Changes


internal/controllers/user_devices_controller.go
```diff
@@ -416,37 +416,47 @@ func (udc *UserDevicesController) GetUserDevices(c *fiber.Ctx) error {
...
+				if udai.Status != "Active" {
+					tok, _ := ud.TokenID.Uint64()
+					toCheck[checkKey{uint32(tok), udai.IntegrationID}] = udai
 				}
+				break
 			}
 		}
 
 		if len(toCheck) != 0 {
 			udc.log.Info().Str("userId", userID).Msgf("Checking %d Ruptela connections.", len(toCheck))
```
zakharenkodmytro, line 442: Should we get rid from Ruptela word, as I understand these changes are generic?

elffjs: Not done

elffjs: Set now

PR #411, opened 2024-11-05
Remove ERC-1271 guards on prod

elffjs:
# Proposed Changes


elffjs:
https://linear.app/dimo/issue/AS-3133/remove-erc-1271-guards-on-prod

PR #419, opened 2024-12-10
AS-3208 - error codes endpoints with tokenID

mateuszpohl:
# Proposed Changes

- `user_device_token_id` in `error_code_queries` (migration, model changes)
- new `/error-codes/*` endpoints using `tokenID` instead of `userDeviceID`
- endpoints require a privileged token with non-location-data privilege
- `/clear` endpoint status codes aligned with the spec (404/429)
- unit tests, new helper

## Note

For now this new field is nullable, since we still may have some devices without a tokenID. Making it not-nullable also breaks other parts (e.g. old version, that we will remove in the future) and older unit tests.


JamesReate:
Ah the `user_device_token_id` we are normally calling `vehicle_token_id`. I know may be pain to change but probably good for consistency...

mateuszpohl:
> Ah the `user_device_token_id` we are normally calling `vehicle_token_id`. I know may be pain to change but probably good for consistency...

changed

internal/controllers/device_data_controller.go
```diff
@@ -276,3 +280,194 @@ func (udc *UserDevicesController) ClearUserDeviceErrorCodeQuery(c *fiber.Ctx) er
...
+// QueryDeviceErrorCodesByTokenID godoc
+// @Summary     Obtain, store, and return descriptions for a list of error codes from this vehicle.
+// @Tags        error-codes
+// @Param       tokenID path int true "vehicle token id"
+// @Param       queryDeviceErrorCodes body controllers.QueryDeviceErrorCodesReq true "error codes"
+// @Success     200 {object} controllers.QueryDeviceErrorCodesResponse
+// @Failure     404 {object} helpers.ErrorRes "Vehicle not found"
+// @Security    BearerAuth
+// @Router      /user/vehicle/{tokenID}/error-codes [post]
+func (udc *UserDevicesController) QueryDeviceErrorCodesByTokenID(c *fiber.Ctx) error {
```
KevinJoiner, line 342: Is this the same function as QueryDeviceErrorCodes other than the ud lookup? If so we can just use a function instead of duplicating the code.

KevinJoiner (changes requested):

internal/controllers/device_data_controller.go
mateuszpohl: These are new ones using vehicle tokenID instead of legacy device ID, the old ones will be removed in future releases, we'll also drop the `user_device_id` column from `error_code_queries` table at that point.

KevinJoiner: Let's mark those as deprecated then

mateuszpohl: Added deprecation info

migrations/20241209201039_add_token_id_to_error_codes.sql
```diff
@@ -0,0 +1,26 @@
...
+-- +goose StatementBegin
+SELECT 'up SQL query';
+SET search_path = devices_api, public;
+ALTER TABLE error_code_queries
+    ADD vehicle_token_id NUMERIC(78, 0);
+
+UPDATE error_code_queries ecq SET vehicle_token_id = ud.token_id FROM user_devices ud WHERE ud.id = ecq.user_device_id;
+
+-- ALTER TABLE error_code_queries
+--     ALTER COLUMN user_device_token_id SET NOT NULL;
```
KevinJoiner, line 11: We can remove this

KevinJoiner (approved):

JamesReate:
This makes me think - this update is mostly for backwards compatibility? like the Ruptela now gets error codes from telemetry-api, and there was discussion of doing same for Macaron/AutoPi? And the chatgpt part / code descriptions would move to a new service.

mateuszpohl:
> This makes me think - this update is mostly for backwards compatibility? like the Ruptela now gets error codes from telemetry-api, and there was discussion of doing same for Macaron/AutoPi? And the chatgpt part / code descriptions would move to a new service.

@JamesReate the main reason for this was (and still is) switch to tokenID instead of legacy vehicle id in the mobile app to finally get rid of the old endpoints. The ultimate solution will be having error codes in telemetry for all HW devices, but there will be some time until we have it.

KevinJoiner (approved):

PR #421, opened 2024-12-10
Tidy

KevinJoiner:
Trying to fix CI

This also removes a call the the device definitions registrar that sent a message to a kafka topic for the device-definition-join-processor to use

### DIFF
--- a/internal/controllers/geofences_controller.go
+++ b/internal/controllers/geofences_controller.go
@@ -275,7 +275,7 @@ func (g *GeofencesController) GetAll(c *fiber.Ctx) error {
 		for _, udtg := range item.R.UserDeviceToGeofences {
 			var deviceDef *ddgrpc.GetDeviceDefinitionItemResponse
 			for _, dd := range dds {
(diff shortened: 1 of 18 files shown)


internal/controllers/geofences_controller.go
```diff
@@ -275,7 +275,7 @@ func (g *GeofencesController) GetAll(c *fiber.Ctx) error {
 		for _, udtg := range item.R.UserDeviceToGeofences {
 			var deviceDef *ddgrpc.GetDeviceDefinitionItemResponse
 			for _, dd := range dds {
-				if dd.DeviceDefinitionId == udtg.R.UserDevice.DeviceDefinitionID {
+				if dd.Id == udtg.R.UserDevice.DefinitionID {
```
mateuszpohl, line 278: I did the same fix on my branch this morning :D

Shouldn't we also add an `if` statement on the part below, just in case? In theory `deviceDef` can still be `nil` for some crazy edge cases.

KevinJoiner: Yeah I think we can just delete this code soon with privacy zones v2 but James is working on fixing up these test. so I'll let him handle it.

PR #427, opened 2025-01-05
Subcommand check-virtual-key to check who has virtual keys

elffjs:
# Proposed Changes


elffjs:
Don't think this test failure is my fault

PR #428, opened 2025-01-09
Update active status check to use multiple sources

KevinJoiner:
This update has the active status check query both the old and the new source in clickhouse.


KevinJoiner:
ref EX-3246

PR #430, opened 2025-01-13
when unpair device with newer endpoint, also mark autopi as unpaired

JamesReate:
# Proposed Changes
- another code flow where autopi's could get marked as unpaired, this stops billing from their end too.

### DIFF
--- a/internal/controllers/user_integrations_controller.go
+++ b/internal/controllers/user_integrations_controller.go
@@ -1511,6 +1511,17 @@ func (udc *UserDevicesController) PostAftermarketDeviceUnpair(c *fiber.Ctx) erro
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


internal/controllers/user_integrations_controller.go
```diff
@@ -1511,6 +1511,17 @@ func (udc *UserDevicesController) PostAftermarketDeviceUnpair(c *fiber.Ctx) erro
 		return err
 	}
 
+	// if this is an autopi, mark it as unpaired on their end so they stop billing us
```
elffjs, line 1514: I don't think anyone calls this anymore. Have you seen any logs?

JamesReate: ugh so much old code in devices-api... I just checked and yeah no calls to that in recent history.

I'll deprecate some stuff instead

PR #435, opened 2025-01-17
EX-3229 Remove "unofficial" Tesla registration code

elffjs:
- Allows us to get rid of TeslaService
- Ignore the version parameter, but still set it in the metadata so we can track

# Proposed Changes


JamesReate (approved):

elffjs:
There may be some limited value in keeping this stuff around while we're still testing out migration stuff.

Going to put this in draft for a bit.

elffjs:
On second thought, we need to remove anything we can.

PR #441, opened 2025-01-30
EX-3119-cio-disconn-event

Allyson-English:
# Proposed Changes


internal/services/cio/cio.go
```diff
@@ -0,0 +1,51 @@
...
+	}, nil
+
+}
+
+func (s *Service) SoftwareDisconnectionEvent(ctx context.Context, vehicleID uint64, address []byte, integrationID string) error {
+	account, err := s.acctClient.GetAccount(ctx, &pb_accounts.GetAccountRequest{
+		WalletAddress: address,
+	})
+	if err != nil {
+		s.logger.Err(err).Str("wallet_address", common.Bytes2Hex(address)).Msg("failed to get account by wallet address from accounts api")
```
KevinJoiner, line 42: should we return here instead of logging
line 45 would panic if account is nil

internal/services/cio/cio.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	return s.client.Enqueue(
+		analytics.Identify{
+			UserId: account.Id,
```
KevinJoiner, line 47: If this is a proto message we should use the function to get the value 
```suggestion
			UserId: account.GetId(),
```

internal/services/task_status_listener.go
```diff
@@ -121,8 +123,16 @@ func (i *TaskStatusListener) processSmartcarPollStatusEvent(event *shared.CloudE
 	}
 
 	udai.Status = models.UserDeviceAPIIntegrationStatusAuthenticationFailure
-	_, err = udai.Update(context.Background(), i.db().Writer, boil.Infer())
-	return err
+	if _, err = udai.Update(context.Background(), i.db().Writer, boil.Infer()); err != nil {
+		i.log.Err(err).Str("userDeviceID", userDeviceID).Str("integrationID", integrationID).Msg("failed up update user device api integration with failure status")
```
KevinJoiner, line 127: Should we be logging or returning the error here.

internal/config/settings.go
```diff
@@ -106,6 +106,9 @@ type Settings struct {
 	DeviceDefinitionsGetByKSUIDEndpoint string `yaml:"DEVICE_DEFINITIONS_GET_BY_KSUID_ENDPOINT"`
 
 	TeslaRequiredScopes string `json:"TESLA_REQUIRED_SCOPES"`
+
+	AccountsAPIGRPCAddr string `json:"ACCOUNTS_API_GRPC_ADDR"`
+	CustomerIOAPIKey    string `yaml:"CUSTOMER_IO_API_KEY"`
```
KevinJoiner, line 111: Do we need a secret to set this value

internal/services/task_status_listener.go
```diff
@@ -121,8 +123,16 @@ func (i *TaskStatusListener) processSmartcarPollStatusEvent(event *shared.CloudE
...
+	}
+
+	vehicleID, ok := udai.R.UserDevice.TokenID.Uint64()
+	if !ok {
+		i.log.Err(errors.New("failed to parse vehicle token id for cio event"))
```
KevinJoiner, line 132: Do we want to return here as well or do we want to send a Disconnection event with out a vehilceId

KevinJoiner (changes requested):

internal/services/task_status_listener.go
Allyson-English: my sense is we should log the error here but keep going. if we aren't able to update the db then the devices api db is out of sync but I don't think that needs to prevent us from sending an update to customer io

KevinJoiner: 👍🏾  I can get behind that since we just log on error any way

internal/services/cio/cio.go
Allyson-English: updated

internal/config/settings.go
Allyson-English: updated

internal/services/task_status_listener.go
Allyson-English: updated to return

charts/devices-api/templates/secret.yaml
```diff
@@ -113,6 +113,9 @@ spec:
   - remoteRef:
       key: {{ .Release.Namespace }}/rewards/clickhouse/pass
     secretKey: CLICKHOUSE_PASSWORD
+  - remoteRef:
```
elffjs, line 116: I'll make some secrets. Need to do this before we merge, I think.

cmd/devices-api/main.go
```diff
@@ -226,7 +230,20 @@ func startTaskStatusConsumer(logger zerolog.Logger, settings *config.Settings, p
 
 	ddSvc := services.NewDeviceDefinitionService(pdb.DBS, &logger, settings)
 
-	taskStatusService := services.NewTaskStatusListener(pdb.DBS, &logger, ddSvc, kp, settings)
+	accountsConn, err := grpc.NewClient(settings.AccountsAPIGRPCAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		logger.Fatal().Err(err).Msg("Failed to create users API client.")
```
elffjs, line 235:
```suggestion
		logger.Fatal().Err(err).Msg("Failed to create accounts API client.")
```

cmd/devices-api/main.go
```diff
@@ -226,7 +230,20 @@ func startTaskStatusConsumer(logger zerolog.Logger, settings *config.Settings, p
...
+	}
+	defer accountsConn.Close()
+
+	acctClient := pb_accounts.NewAccountsClient(accountsConn)
+
+	cioSvc, err := cio.New(settings.CustomerIOAPIKey, acctClient, &logger)
+	if err != nil {
+		logger.Fatal().Err(err).Msg("failed to start customer io service")
```
elffjs, line 243:
```suggestion
		logger.Fatal().Err(err).Msg("failed to create customer io service")
```
This is picky but it's not, like, a daemon in my view.

internal/services/task_status_listener.go
elffjs: Our error handling when Kafka messages fail to process is... it's not good. I don't see a need to _change_ the behavior like we're doing here but

internal/services/cio/cio.go
```diff
@@ -0,0 +1,52 @@
...
+	})
+	if err != nil {
+		s.logger.Err(err).Str("wallet_address", common.Bytes2Hex(address)).Msg("failed to get account by wallet address from accounts api")
+		return err
+	}
+
+	return s.client.Enqueue(
+		analytics.Identify{
+			UserId: account.GetId(),
+			Traits: analytics.NewTraits().Set("integration_id", integrationID).Set("vehicle_id", vehicleID),
```
elffjs, line 49: I think the integration id should be the number—(1 for Smartcar, 2 for Tesla). Trying to leave these KSUIDs behind!

I think we also want—and I'm sorry this is buried in a comment—the "device id"

https://linear.app/dimo/issue/EX-3119/cio-event-for-software-disconnections#comment-bf757a89

I take this to mean the synthetic device's address, which we might as well checksum

elffjs (changes requested):

charts/devices-api/templates/secret.yaml
elffjs: Done.

elffjs (changes requested):

internal/services/cio/cio.go
```diff
@@ -0,0 +1,91 @@
...
+	if !userMixAddr.ValidChecksum() {
+		return fmt.Errorf("invalid ethereum_address %s", common.BytesToAddress(udai.R.UserDevice.OwnerAddress.Bytes))
+	}
+
+	sd := udai.R.UserDevice.R.VehicleTokenSyntheticDevice
+	if sd == nil {
+		return errors.New("no synthetic device associcated with api integration")
+	}
+
+	sdWallet := common.NewMixedcaseAddress(common.BytesToAddress(sd.WalletAddress))
```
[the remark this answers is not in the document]
Allyson-English: ah, yeah that's what I would ordinarily do but I interpreted your comment before about doing a checksum as wanting to do a mixed addr and then call `.ValidChecksum()

internal/services/cio/cio.go
```diff
@@ -0,0 +1,83 @@
...
+	return &Service{
+		client:     client,
+		acctClient: acctClient,
+		logger:     logger,
+	}, nil
+
+}
+
+func (s *Service) SoftwareDisconnectionEvent(ctx context.Context, udai *models.UserDeviceAPIIntegration) error {
+	vehicleTokenID, ok := udai.R.UserDevice.TokenID.Int64()
```
elffjs, line 43: Let's make sure that this thing has a token id in the first place. There are probably still some vestigial rows.

internal/services/cio/cio.go
```diff
@@ -0,0 +1,83 @@
...
+	if !ok {
+		return errors.New("failed to parse vehicle token id")
+	}
+
+	userAddr := common.BytesToAddress(udai.R.UserDevice.OwnerAddress.Bytes)
```
elffjs, line 48: Similarly, maybe there's no owner.

internal/services/cio/cio.go
```diff
@@ -0,0 +1,83 @@
...
+	if !common.IsHexAddress(userAddr.Hex()) {
+		return fmt.Errorf("invalid ethereum_address %s", userAddr.Hex())
+	}
```
elffjs, line 51: I think there's no danger of this happening. If you wanted to be really, really careful you could check that `udai.R.UserDevice.OwnerAddress.Bytes` really has length 20, but I think the database will enforce that if it's populated so I think it's not too scary.

Allyson-English: okay cool, removed it in that case

internal/services/cio/cio.go
```diff
@@ -0,0 +1,87 @@
...
+
+	sd := udai.R.UserDevice.R.VehicleTokenSyntheticDevice
+	if sd == nil {
+		return errors.New("no synthetic device associcated with api integration")
+	}
+
+	sdWallet := common.BytesToAddress(sd.WalletAddress)
+	if !common.IsHexAddress(sdWallet.Hex()) {
+		return fmt.Errorf("invalid ethereum_address %s", sdWallet.Hex())
+	}
```
KevinJoiner, line 64: `sdWallet.Hex()` will always produce a valid address so this check won't ever get hit. Maybe we can just check if it is nil
https://go.dev/play/p/GmCQRL7k5-v

internal/services/task_status_listener.go
```diff
@@ -121,8 +125,16 @@ func (i *TaskStatusListener) processSmartcarPollStatusEvent(event *shared.CloudE
...
 
 	udai.Status = models.UserDeviceAPIIntegrationStatusAuthenticationFailure
-	_, err = udai.Update(context.Background(), i.db().Writer, boil.Infer())
-	return err
+	if _, err = udai.Update(context.Background(), i.db().Writer, boil.Infer()); err != nil {
+		i.log.Err(err).Str("userDeviceID", userDeviceID).Str("integrationID", integrationID).Msg("failed up update user device api integration with failure status")
+	}
+
+	if err := i.cioSvc.SoftwareDisconnectionEvent(ctx, udai); err != nil {
+		i.log.Err(err).Str("userDeviceID", userDeviceID).Str("integrationID", integrationID).Msg("failed to send CIO software disconnection event")
```
KevinJoiner, line 133: I think this error will be logged by the caller so we don't need to log here as well.

internal/services/task_status_listener.go
```diff
@@ -167,9 +180,16 @@ func (i *TaskStatusListener) processTeslaPollStatusEvent(event *shared.CloudEven
 		}
 	}
 	udai.Status = models.UserDeviceAPIIntegrationStatusAuthenticationFailure
+	if _, err = udai.Update(context.Background(), i.db().Writer, boil.Infer()); err != nil {
+		i.log.Err(err).Str("userDeviceID", userDeviceID).Str("integrationID", integrationID).Msg("failed up update user device api integration with failure status")
+	}
+
+	if err := i.cioSvc.SoftwareDisconnectionEvent(ctx, udai); err != nil {
+		i.log.Err(err).Str("userDeviceID", userDeviceID).Str("integrationID", integrationID).Msg("failed to send CIO software disconnection event")
```
KevinJoiner, line 188: Same as above we don't want to error logs for the same error

KevinJoiner (approved):
small suggestions overall looks good

elffjs (approved):
Would like a small deletion

PR #445, opened 2025-02-05
ex-3332-add-sacd-parameters-to-minting-endpoints

Allyson-English:
# Proposed Changes
- Adds SACDInput to mint vehicle request body and includes this as part of the mint vehicle with synthetic device call

### Impacted Routes
new endpoint
- `udOwner.Post("/commands/mint", userDeviceController.PostMintDevice)

### Notes
- This doesn't add an SACD when we are just minting a synthetic device because permissions are based on vehicle not integration


internal/controllers/user_devices_controller.go
```diff
@@ -1668,7 +1917,8 @@ func (udc *UserDevicesController) UpdateNFTImage(c *fiber.Ctx) error {
 type VehicleMintRequest struct {
 	NFTImageData
 	// Signature is the hex encoding of the EIP-712 signature result.
-	Signature string `json:"signature" validate:"required"`
+	Signature string               `json:"signature" validate:"required"`
+	SACDInput *contracts.SacdInput `json:"sacd_input,omitempty"`
```
KevinJoiner, line 1921: maybe use separate  struct so we can set JSON tags

cmd/devices-api/api.go
```diff
@@ -239,6 +239,7 @@ func startWebAPI(logger zerolog.Logger, settings *config.Settings, pdb db.Store,
 	udOwner.Delete("/", userDeviceController.DeleteUserDevice)
 	udOwner.Get("/commands/mint", userDeviceController.GetMintDevice)
 	udOwner.Post("/commands/mint", userDeviceController.PostMintDevice)
+	udOwner.Post("/commands/mint-with-sacd", userDeviceController.PostMintDeviceWithSACD)
```
KevinJoiner, line 242: wombo combo these bad boys

internal/controllers/user_devices_controller.go
KevinJoiner, line 1921:
```suggestion
	SACDInput *contracts.SacdInput `json:"sacdInput,omitempty"`
```

KevinJoiner (changes requested):

internal/controllers/user_devices_controller.go
elffjs, line 1922: Let's add a little documentation here. We will certainly have to regenerate Swagger.

cmd/devices-api/api.go
elffjs: I, too, think it's okay to keep the same endpoint.

Not that you couldn't do this even with your setup, but I think it would also encourage re-using the code.

internal/controllers/user_devices_controller.go
elffjs:
```
+  github_com_DIMO-Network_devices-api_internal_contracts.SacdInput:
+    properties:
+      expiration:
+        $ref: '#/definitions/big.Int'
+      grantee:
+        items:
+          type: integer
+        type: array
+      permissions:
+        $ref: '#/definitions/big.Int'
+      source:
+        type: string
+    type: object
```
The names actually turned out okay! But the types are kinda fried.

It might be a worthwhile follow-up to do a little validation on these guys.

elffjs (changes requested):
We're close.

cmd/devices-api/api.go
Allyson-English: yep, I agree-- the original ticket said to do it in two different endpoints, which is why I took this approach initially

KevinJoiner (changes requested):

internal/contracts/registry.go
```diff
@@ -1958,6 +2190,27 @@ func (_Registry *RegistryTransactorSession) MintVehicleAndSdWithDeviceDefinition
 	return _Registry.Contract.MintVehicleAndSdWithDeviceDefinitionSign(&_Registry.TransactOpts, data)
 }
 
+// MintVehicleAndSdWithDeviceDefinitionSignWithSacd is a paid mutator transaction binding the contract method 0x62a147ff.
```
KevinJoiner, line 2193: I think this needs to be regenerated with the new name mintVehicleAndSdWithDeviceDefinitionSignAndSacd

internal/controllers/user_devices_controller.go
```diff
@@ -1670,6 +1690,19 @@ type VehicleMintRequest struct {
...
 	Signature string `json:"signature" validate:"required"`
+	// SACDInput contains user signed permission grant, including grantee, permissions, expiration and link to signed document
+	SACDInput *SACDInput `json:"sacdInput,omitempty"`
+}
+
+type SACDInput struct {
+	// Grantee is the Ethereum address permissions are being granted to.
+	Grantee common.Address `json:"grantee"`
+	// Permissions are a numerical representation of what permissions are being given to the grantee.
+	Permissions *big.Int `json:"permissions"`
```
elffjs, line 1701: I am worried about how this will look in Swagger. Can we run this?

elffjs (approved):
Let's just try it out.

PR #478, opened 2025-03-26
Decode Tesla VINs with DD-api

JamesReate:
# Proposed Changes

main advantage of doing this is that the definition will get created if it doesn't exist

underlying logic in DD-api has been changed to not decode with a vendor but just use same logic as had here, so failure rate should be much lower.

### DIFF
--- a/internal/controllers/user_integrations_auth_controller.go
+++ b/internal/controllers/user_integrations_auth_controller.go
@@ -252,6 +252,6 @@ func (u *UserIntegrationAuthController) decodeTeslaVIN(ctx context.Context, vin 
 
 	teslaMake := "Tesla"
 	model := shared.VIN(vin).TeslaModel()
-	// decode vin doesn't return Model, just the ID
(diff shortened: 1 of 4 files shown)


internal/controllers/user_integrations_auth_controller.go
```diff
@@ -244,14 +244,14 @@ type decodeResult struct {
...
+	// for Tesla, this does not call vendor to decode, uses same logic as below - advantage is it will create the DD if it doesn't exist
+	decodeVIN, err := u.DeviceDefSvc.DecodeVIN(ctx, vin, "", 0, "USA")
 	if err != nil {
 		return nil, err
 	}
 
-	return &decodeResult{ID: res.DeviceDefinitionId, Make: teslaMake, Model: model, Year: year}, nil
+	teslaMake := "Tesla"
+	model := shared.VIN(vin).TeslaModel()
+	// decode vin doesn't return Model, just the ID
```
elffjs, line 255: Should we just call DD again...?

elffjs (approved):

PR #514, opened 2025-08-07
PR-3803: Stop tesla cmd when subscription status not active

zakharenkodmytro:
- updated tesla-oracle dependency
- updated shared library and all imports
-  drop commands request if subscription is not active


elffjs:
I would like to have separated out 

1. This big dependency upgrade 
2. The feature work

zakharenkodmytro:
> I would like to have separated out
> 
> 1. This big dependency upgrade
> 2. The feature work

Yeah, that what I was thinking but then realized it is hard to do

PR #516, opened 2025-08-09
Really delete Smartcar, and some other stuff

elffjs:
* Get rid of all Smartcar handling code. I left a few display-only blocks in there, just in case something's hanging around in device-definitions-api that we might be forced to serve. Stop showing it in our hardcoded compatibility list.
* Remove `/fromsmartcar`, `/fromvin`, `/refresh`. [No calls to these in a long time.](https://kibana.team.dimo.zone/app/r/s/YHaa8)
* Remove the event emitter. No one uses this anymore; use the contract events instead.
* Remove some unused settings.
* Move a synthetic device test from Smartcar to Tesla; no more magic strings.

I think there's more that could be cut: 

* I think there is more aftermarket code here that doesn't serve a purpose anymore.
* I think some of the S3 stuff could also go.
* Some Kafka topics could be removed.

### DIFF
--- a/cmd/devices-api/set_command_compat.go
+++ b/cmd/devices-api/set_command_compat.go
@@ -16,11 +16,10 @@ import (
 )
 
 type setCommandCompatibilityCmd struct {
-	logger    zerolog.Logger
-	settings  config.Settings
-	pdb       db.Store
-	ddSvc     services.DeviceDefinitionService
-	container dependencyContainer
+	logger   zerolog.Logger
+	settings config.Settings
+	pdb      db.Store
+	ddSvc    services.DeviceDefinitionService
 }
 
 func (*setCommandCompatibilityCmd) Name() string     { return "set-command-compat" }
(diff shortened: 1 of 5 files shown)


internal/controllers/synthetic_devices_controller.go
```diff
@@ -40,7 +40,11 @@ type SyntheticDevicesController struct {
 	deviceDefSvc   services.DeviceDefinitionService
 	walletSvc      services.SyntheticWalletInstanceService
 	registryClient registry.Client
-	teslaOracle    pb_oracle.TeslaOracleClient
+	teslaOracle    TeslaOracleClient
+}
+
+type TeslaOracleClient interface {
```
zakharenkodmytro, line 46: Agree, better design approach

elffjs: Eh it's mostly so that I can mock it in the test.

JamesReate (approved):
let's try this out!

